Issue #107, opened 2014-07-31
Standalone installation of Linuxbrew


mbsullivan:
I just tested the installation of glibc under the bottle branch; it went pretty well, but there may be a couple missing dependencies. (I am installing it on a machine running CentOS 5.5, which is quite outdated.)

I haven't yet looked into how dependencies/version requirements are dictated in Homebrew/Linuxbrew, so I'll just describe the procedure I went through to get it installed. Here's what happened:

(1) prepared the bottle branch and installed patchelf using Linuxbrew's local gcc:
```
>cd $LINUXBREW; git checkout bottle
>brew install patchelf -v --cc=gcc-4.8
```

(2) tried glibc `brew install glibc -v --cc=glibc`, it fails with:
```
*** These critical programs are missing or too old: as ld
```
It seems to me like this means that a new version of binutils is a dependency.

(3) installed binutils `brew install binutils -v --cc=gcc-4.8`. 

(4) The binutils [recipe](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/binutils.rb) uses a program prefix 'g' (leading to `gld` and `gas`), which is non-standard for Linux distros I've used. This led to some errors, so I had to manually symlink $LINUXBREW/bin/ld and $LINUXBREW/bin/gas to the generated programs. Would this be considered a bug in linuxbrew's binutils?

(5) With the new binutils, an installation of glibc now runs into old kernel header files:
```
checking installed Linux kernel header files... TOO OLD!
```
It seems like this means that a new version of linux-headers is a dependency.

(4) Installed linux-headers with `brew install linux-headers -v --cc=gcc-4.8`. Something weird happened here... It seemed to not respect the --cc= version I gave it, and ended up using an old-and-busted gcc on the system path. (Any idea why this would happen?) I took this version of gcc out of my path and continued.

(5) With new binutils and linux-headers installed, glibc 2.19 installed like a charm. 

One more thing: I might add a caveat to the installer to warn the user to unset LD_LIBRARY_PATH immediately if they've left it pointing to $LINUXBREW/lib/. Otherwise the new glibc will minorly hose your system if it expects an older one.

yeluolei:
@mbsullivan same problem with an old gcc version 3.5 and no system access rights, any process to install a new gcc?

sjackman:
I believe these issues are resolved. I'm closing this ticket. Feel free to open a new ticket if some of the issues raised here are still current.

PR #200, opened 2014-12-07
gcc, binutils: link to Cellar instead of system libs

jbaum98:
Symlinks `linuxbrew/lib` to Cellar for `gcc`, where `gcc` will automatically look because that is the prefix. Should fix the issue discussed recently in #107.

### DIFF
--- a/Library/Formula/gcc.rb
+++ b/Library/Formula/gcc.rb
@@ -166,6 +166,12 @@ class Gcc < Formula
         args << "--with-sysroot=#{MacOS.sdk_path}"
       end
 
+      if OS.linux?
+        @link = Pathname.new "#{prefix}/x86_64-unknown-linux-gnu/bin"
+        @link.parent.mkpath
+        @link.make_symlink "#{HOMEBREW_PREFIX}/lib"
+      end
+
       system "../configure", *args
       system "make", "bootstrap"
       system "make", "install"
@@ -210,6 +216,17 @@ class Gcc < Formula
       rmdir lib64
       prefix.install_symlink "lib" => "lib64"
     end
+    
+    if OS.linux?
+      p = @link.parent
+      @link.delete
+      p.delete
+      crts = Pathname.new "#{lib}/gcc/x86_64-unknown-linux-gnu/#{version}"
+      Formula['glibc'].lib.children.select {|p| p.basename.to_s =~ /^crt.\.o$/ }.collect {|p| p.relative_path_from crts}.each do |p|
+        crts.install_symlink p 
+      end
+    end
+
   end
 
   def add_suffix file, suffix
--- a/Library/Formula/binutils.rb
+++ b/Library/Formula/binutils.rb
@@ -26,6 +26,7 @@ class Binutils < Formula
                           "--disable-werror",
                           "--enable-interwork",
                           "--enable-multilib",
+                          ("--with-lib-path=#{HOMEBREW_PREFIX}/lib" if OS.linux?),
                           "--enable-64-bit-bfd",
                           "--enable-targets=all"
     system "make"


Library/Formula/gcc.rb
```diff
@@ -210,6 +216,17 @@ def install
...
+    
+    if OS.linux?
+      p = @link.parent
+      @link.delete
+      p.delete
+      crts = Pathname.new "#{lib}/gcc/x86_64-unknown-linux-gnu/#{version}"
+      Formula['glibc'].lib.children.select {|p| p.basename.to_s =~ /^crt.\.o$/ }.collect {|p| p.relative_path_from crts}.each do |p|
+        crts.install_symlink p 
+      end
+    end
```
sjackman: I understand that this patch allows GCC to find crt1.o, crti.o and crtn.o by linking them into a directory that GCC searches by default. What's the purpose of the above symlink from `"#{HOMEBREW_PREFIX}/lib"` to `"#{prefix}/x86_64-unknown-linux-gnu/bin"`?

jbaum98: xgcc automatically looks in that location for certain lib files. It is related to the -B and GCC_EXEC_PREFIX. I couldn't figure out how to pass command line flags to xgcc so that was how I made those files available to it.

davydden:
so it seems that `gfortran` issue is still open on a stand-alone installation?..

bobobo1618:
So I managed to build GCC with this patch but attempting to compile binaries using the new GCC results in the following error:

```
Error: undefined method `gcc-4.9_build_version' for OS::Mac:Module
Please report this bug:
    https://github.com/Homebrew/linuxbrew/blob/master/share/doc/homebrew/Troubleshooting.md#troubleshooting
/media/drive/home/user/.linuxbrew/Library/Homebrew/compilers.rb:119:in `block in find_compiler'
/media/drive/home/user/.linuxbrew/Library/Homebrew/compilers.rb:106:in `each'
/media/drive/home/user/.linuxbrew/Library/Homebrew/compilers.rb:106:in `find_compiler'
/media/drive/home/user/.linuxbrew/Library/Homebrew/compilers.rb:99:in `compiler'
/media/drive/home/user/.linuxbrew/Library/Homebrew/compilers.rb:82:in `select_for'
/media/drive/home/user/.linuxbrew/Library/Homebrew/extend/ENV/shared.rb:123:in `compiler'
/media/drive/home/user/.linuxbrew/Library/Homebrew/extend/ENV/std.rb:74:in `setup_build_environment'
/media/drive/home/user/.linuxbrew/Library/Homebrew/build.rb:90:in `install'
/media/drive/home/user/.linuxbrew/Library/Homebrew/build.rb:179:in `<main>'
```

I simply edited `compilers.rb`, took out the switch statement and hard coded it to the `:gnu` case and it works. That seems terribly hacky though and I'm sure there's something up. Shouldn't `:gnu` be triggered anyway?

iggyvolz:
Looks like #404 where it ignores HOMEBREW_CC.  Using --cc=gcc4.9 should fix the problem

sjackman:
This bug should be fixed upstream by Homebrew/homebrew@43c634f, which has been merged.

ahundt:
Can issues discussed here be related to the problems I'm encountering in #530? I may also be encountering the `crtn.o` issues discussed here.

sjackman:
@jrevillard I believe that 9318c13512d592db39e9bc29532bc4317ec8733c should fix the `gfortran` issue.
@jbaum98 Fixing the search path for `crti.o` and friends when linking other programs should be possible by adding `-B#{HOMEBREW_PREFIX}/lib` to `LDFLAGS` in `ENV/std.rb`.

I'm going to close this PR. Thanks for all your help in troubleshooting. If you still have current issues, please do comment or open a new issue to discuss them.

PR #217, opened 2014-12-23
valgrind 3.10.1: Fixes for Linuxbrew

jlisic:
Conditioned the flag for `--build=amd64-darwin` on the host being OS X, otherwise the formula fails.  A few  OS X version checks were also conditioned on the OS being OS X, this is mainly for aesthetic reasons.


sjackman:
See #226

PR #228, opened 2015-01-02
Fix for java dependency check on linux

saketkc:
See: https://github.com/Homebrew/homebrew-science/issues/1605#issuecomment-68510052

PR #229, opened 2015-01-04
JavaDependency: Fix for java version checks

saketkc:
Related #228

### DIFF
--- a/Library/Homebrew/requirements.rb
+++ b/Library/Homebrew/requirements.rb
@@ -123,7 +123,8 @@ class JavaDependency < Requirement
   end
 
   def java_version
-    return quiet_system "java", "-version" if OS.linux?
+    args = %W[java -version 2>&1 | awk -F '"' '/version/ {print ($2 >= "#{@version}")?err=0:err=1} END{exit err}']
+    quiet_system(*args) if OS.linux?
     args = %w[/usr/libexec/java_home --failfast]
     args << "--version" << "#{@version}+" if @version
     quiet_system(*args)


sjackman:
Could you instead parse the output of `java -version` using Ruby rather than awk?

Until Linuxbrew supports installing multiple versions of JDK, I'm not sure that we want to be super fussy about checking that the version is correct. Are the formula that definitely require a particular version of JDK to work?

saketkc:
Not a rubyist, but I am going to give it a short.
Not sure about any formula asserting `jdk` version to be above certain version.

saketkc:
@sjackman I am not sure if the ruby version can be tidied up. However please not that you cannot use `return quiet_system whatever`, else it would pass through, always(Since it always returns non null as opposed to just using `quiet_system` without `return`).

sjackman:
@saketkc I was thinking someting more like
```ruby
`java -version 2>&1`.include? "1.8"
```

sjackman:
I just pushed a `jdk7` formula.

saketkc:
@sjackman You would be able to check only for a particular version using ``java -version 2>&1`.include? "1.8"`. What if 1.8+  is supported, like 1.9.0_67 ?

sjackman:
I don't believe Homebrew has support for that kind of minimum (or maximum) version check with Java.

saketkc:
Sorry, my bad. I must have misunderstood the check to verify a minimal version, since the Mac specific code checked for `1.7+`.

sjackman:
Ah, no. It's my mistake! I missed the all important plus sign in `args << "--version" << "#{@version}+" if @version`. It should be a minimal version.

```
~ ❯❯❯ /usr/libexec/java_home --failfast --version 1.8
/Library/Java/JavaVirtualMachines/jdk1.8.0_25.jdk/Contents/Home
~ ❯❯❯ /usr/libexec/java_home --failfast --version 1.7
Unable to find any JVMs matching version "1.7".
~ ❯❯❯ /usr/libexec/java_home --failfast --version 1.7+
/Library/Java/JavaVirtualMachines/jdk1.8.0_25.jdk/Contents/Home
```

saketkc:
Good to merge @sjackman?

sjackman:
I'd prefer that the awk script be implemented in Ruby. Something like (untested)…
```ruby
`java -version 2>&1`[/java version "(.*)"/,1].to_f >= (@version || 1) if OS.linux?
```

saketkc:
I did it at one point of time. Looks like I forced rewrite my commits.

saketkc:
Found it: https://github.com/saketkc/linuxbrew/commit/e9104d5e027345c82885915b50bfac9829a219fe

sjackman:
Piping into ruby is quite unnecessary. You're already running in a Ruby interpreter? Does my one-line version above work?

PR #231, opened 2015-01-07
p7zip: Fix for linuxbrew

saketkc:


sjackman:
~~~
+brew audit p7zip
p7zip:
 * Use the `chmod` Ruby method instead of `system "chmod `
Error: 1 problems in 1 formulae
The command "brew test-bot --dry-run $TRAVIS_COMMIT |sed -n 's/^==> //p' |sh -ex" exited with 1.
~~~

sjackman:
The patch can be made smaller by using the suffix form `end if OS.mac?`. It makes merging conflicts easier.

saketkc:
@sjackman If I use `chmod 776, "#{doc}"` I get:
```
==> make DEST_HOME=/home/saket/.linuxbrew/Cellar/p7zip/9.20.1 DEST_MAN=/home/saket/.linuxbrew/Cellar/p7zip/9.20.1/share/man install
find: `/home/saket/.linuxbrew/Cellar/p7zip/9.20.1/share/doc/p7zip': Permission denied
/usr/bin/du: cannot access ‘/home/saket/.linuxbrew/Cellar/p7zip/9.20.1/share/doc/p7zip/ChangeLog’: Permission denied
/usr/bin/du: cannot access ‘/home/saket/.linuxbrew/Cellar/p7zip/9.20.1/share/doc/p7zip/README’: Permission denied
/usr/bin/du: cannot access ‘/home/saket/.linuxbrew/Cellar/p7zip/9.20.1/share/doc/p7zip/DOCS’: Permission denied

```
Is this not the right way to it?

sjackman:
Try `chmod 0776, doc`. The `0` indicates octal format.

Why do you need to change the perms? Is there an error message? The perms should probably be `755`.

sjackman:
You can simply delete that line `system "chmod -R +w #{doc}"`. It's not needed anymore. Please add a `test do end` if you have the time.

saketkc:
I edited the `install.sh` script to get it working.

sjackman:
You're right. I messed up my test (probably installed the bottle by mistake). The following works as well. I prefer this solution because it's smaller.
```ruby
    # install.sh chmods to 444, which is bad and breaks uninstalling
    chmod_R "u+w", doc
```

Library/Formula/p7zip.rb
```diff
@@ -18,7 +18,13 @@ def install
       mv 'makefile.macosx_32bits', 'makefile.machine'
     else
       mv 'makefile.macosx_64bits', 'makefile.machine'
-    end
+    end if OS.mac?
+
+    mv 'makefile.linux_any_cpu', 'makefile.machine' if OS.linux?
+
+    # install.sh chmods to 444, which is bad and breaks uninstalling
+    inreplace "install.sh", "chmod 444", "chmod 0775"
```
sjackman: Files should not be executable. `0775` should be `0644`.

sjackman:
I take it back. My preference is simply to delete the two `chmod` lines. That should work (untested).

saketkc:
Fixed.

Library/Formula/p7zip.rb
```diff
@@ -18,16 +18,19 @@ def install
...
       mv 'makefile.macosx_64bits', 'makefile.machine'
-    end
+    end if OS.mac?
+
+    mv 'makefile.linux_any_cpu', 'makefile.machine' if OS.linux?
 
+    # install.sh chmods to 444, which is bad and breaks uninstalling
+    inreplace "install.sh" do |s|
+      s.gsub! /chmod 444.*|chmod 555.*/, ""
+    end
```
sjackman: Tidier (untested)
```ruby
inreplace "install.sh", /^chmod .*/, ""
```

sjackman: Can you submit this chunk (the non-Linux-specific stuff) to Homebrew/homebrew?

saketkc: `inreplace "install.sh", /^chmod .*/, ""` would be too general, because there are lines like:

`chmod 555 "${DEST_DIR}${DEST_SHARE}/${prg}"`

saketkc:
Submitted https://github.com/Homebrew/homebrew/pull/35675/

sjackman:
```ruby
inreplace "install.sh", /^chmod (444|555).*/, ""
```

sjackman:
Great. Thanks.

saketkc:
`inreplace "install.sh", /^chmod (444|555).*/, ""`
Makes sense, will do.

PR #245, opened 2015-01-15
Explicit symlink instructions for Fedora/Centos

dctrud:
Prominent, explicit symlink instructions for Fedora/CentOS/Redhat. brew doesn't work without these symlinks.


sjackman:
Woops. I read #246 before I read this PR. Thanks for the patch. In any case, hopefully it's fixed now by 63b1cd37d84e6d1ce57d8a118edfa77e05d66162.

PR #246, opened 2015-01-15
Add explicit Fedora/CentOS symlink to gh-pages

dctrud:


sjackman:
`index.html` is generated automatically from `README.md`. This patch should modify the latter. No need though, because it gave me the encouragement to actually fix Linuxbrew. In addition to searching for compilers named like `gcc-4.9`, it now also searches for compilers named like `gcc44` and simply `gcc`.

Fixed in 63b1cd37d84e6d1ce57d8a118edfa77e05d66162

PR #249, opened 2015-01-15
.travis.yml: disable emails & enable fast_finish

### LINKED ISSUE #248: Please disable Linuxbrew's Travis from emailing Homebrew committers
Thanks!

Mikaela:
Closes #248. I am not aware of any other way to fix it.

Fast_finish makes the whole build as failed if one build fails instead
of having to wait for all builds to finish. It's probably not needed
yet, but it don't hurt to have there and it's probably possible that
there can be more builds in the future.


Mikaela:
The previous sync was to fix Windows line-ending which got there earlier.

sjackman:
Thanks for the patch. I didn't pull the fast finish until I find a use case to support it.

PR #251, opened 2015-01-17
Fix dependencies for pcre

paulhybryant:
Fix dependencies for pcre


sjackman:
These core dependencies are expected to be provided by the host operating system for a standard install of Linuxbrew. I haven't yet sorted out how to deal with this issue for Standalone Linuxbrew, other than to suggest installing all formula found in Homebrew/dupes.

PR #255, opened 2015-01-22
xrootd: no need to fix manpath on linux (GNUInstallDirs is used)

veprbl:
The install paths handling is different on linux and it's already homebrew complaint so we don't have to fix it:
https://github.com/xrootd/xrootd/blob/4d465799a237523651883faf8f4b09b09d5918d5/cmake/XRootDOSDefs.cmake#L32-L55

### DIFF
--- a/Library/Formula/xrootd.rb
+++ b/Library/Formula/xrootd.rb
@@ -20,7 +20,7 @@ class Xrootd < Formula
       system "cmake", "..", *std_cmake_args
       system "make", "install"
     end
-    share.install prefix/"man"
+    share.install prefix/"man" unless OS.linux?
   end
 
   test do


sjackman:
On Mac OS `man` should not be installed in prefix. This smells like an upstream bug. Can you report this bug upstream and include a link to the issue here?

`if not` should be `unless`

Library/Formula/xrootd.rb
```diff
@@ -20,7 +20,7 @@ def install
       system "cmake", "..", *std_cmake_args
       system "make", "install"
     end
-    share.install prefix/"man"
+    share.install prefix/"man" if not OS.linux?
```
sjackman: `if not` should be `unless`

veprbl:
Updated.

veprbl:
https://github.com/xrootd/xrootd/issues/193

sjackman:
Thanks for the fix!

PR #262, opened 2015-01-26
vim 7.4.488: Fix for Linuxbrew

matope:
Hi.
In my centos 6.5, ``brew install vim`` fails due to missing of link to `ncurses`.
When I added ``homebrew/dupes/ncurses`` to depends_on, install succeeded.
So I'd like to send patch, but I'm not sure if this is a correct way to solve it (Can I specify dupes formula as a depends_on ?).
Could you please review this patch?

### DIFF
--- a/Library/Formula/vim.rb
+++ b/Library/Formula/vim.rb
@@ -32,6 +32,7 @@ class Vim < Formula
   depends_on "lua" => :optional
   depends_on "luajit" => :optional
   depends_on "gtk+" if build.with? "client-server"
+  depends_on "homebrew/dupes/ncurses" => :build
 
   conflicts_with "ex-vi",
     :because => "vim and ex-vi both install bin/ex and bin/view"


sjackman:
The current policy is that Linuxbrew must install all the formula provided by `homebrew/dupes` to provide a system that has all the dependencies that are usually provided by Mac OS.

sjackman:
Thanks for the feedback though. The above policy may change.

matope:
My understanding is that the policy for dependencies on ``homebrew/dupes`` has not been  determined yet and if I would like to pass through this issue, I should have install all ``homebrew/dupes`` manually. 

OK. I see. I close this ticket. I hope the policy will be determined nicely.

sjackman:
Yes, you understand correctly. Thanks for your understanding. I'd like to eventually add dependencies like `depends_on :ncurses` that on Mac is satisfied by Mac OS and on Linux possibly checks whether a native package is installed, and otherwise installs `homebrew/dupes/ncurses`.

PR #263, opened 2015-01-26
mutt: Fix for linuxbrew

saketkc:

PR #264, opened 2015-01-26
travisCI: Install pkg-config

saketkc:
See build results for #263

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ install:
  # Incorrect file permissions: chmod 644
  - chmod 644 Library/Formula/*.rb
  - PATH=$PWD/bin:$PATH
+ # Install pkg-config
+ - brew install pkg-config
 script:
  - brew test-bot --dry-run $TRAVIS_COMMIT |sed -n 's/^==> //p' |sh -ex
 notifications:


saketkc:
Good to merge? @sjackman

sjackman:
I'd prefer to understand why packages that depend on pkg-config aren't installing it correctly. Are you able to do some troubleshooting into the root cause?

saketkc:
Changing `depends_on 'pkg-config => :build` in `cloog.rb` to a hard dependency?

sjackman:
@saketkc Do you mean get rid of `=> :build`? This bug finally came up on Docker:
https://registry.hub.docker.com/u/sjackman/linuxbrew-bottle/build_id/22230/code/be6fqn4fpychd89sahnraqw/
```
==> Installing gcc dependency: cloog
Error: /home/linuxbrew/.linuxbrew/opt/pkg-config not present or broken
Please reinstall pkg-config. Sorry :(
```

saketkc:
Yes, I think we should get rid of `:build`

sjackman:
Just for `cloog`? All the other formula that depend on `pkg-config` use `=> :build`. Why would only `cloog` be broken? I think this is a Homebrew bug, but I haven't been able to figure it out.

lpar:
I just got this trying to install neovim:

    ==> Installing neovim/neovim/neovim dependency: makedepend
    Error: /home/meta/.linuxbrew/opt/pkg-config not present or broken

sjackman:
Hi, Matthew. Does `brew install pkg-config` before `brew install neovim` resolve your issue?

sjackman:
Oh! Revelation, @saketkc! I found the root cause of this bug. When Homebrew does its dependency analysis it determines that it will install a *bottle* of `makedepend`, because there are Mac bottles available. When it instead installs `makedepend` by source, it finds that `pkg-config` is not installed. It didn't install `pkg-config`, because `pkg-config` is not required by `makedepend` bottle.

The real fix would be for the dependency logic to check that a bottle is available *for this architecture*.

@xu-cheng Any thoughts on this bug?

xu-cheng:
> When it instead installs makedepend by source, it finds that pkg-config is not installed. It didn't install pkg-config, because pkg-config is not required by makedepend bottle.

That's unlikely as `pkg-config` should be installed when installing `makedepend` from source.

> The real fix would be for the dependency logic to check that a bottle is available for this architecture

This logic exists for all along. 

Is there a minimal example to reproduce this bug so I can debug? Also @jacknagel should know more  than me about dependency issues.

jacknagel:
I would need more detailed build output and some steps to reproduce, otherwise I can only guess.

lpar:
```
% brew install pkg-config
Error: pkg-config cannot be built with any available compilers.
To install this formula, you may need to:
  brew install gcc
% gcc --version
gcc (GCC) 4.9.2 20150212 (Red Hat 4.9.2-6)
% which pkg-config
/usr/bin/pkg-config
% pkg-config --version
0.28
```

lpar:
Is this minimal enough?
```
% brew install -v bug264.rb 
==> Installing bug264 dependency: makedepend
Error: /home/meta/.linuxbrew/opt/pkg-config not present or broken
Please reinstall pkg-config. Sorry :(
% cat bug264.rb 
class Bug264 < Formula
  homepage "https://github.com/Homebrew/linuxbrew/pull/264"
  url "https://raw.githubusercontent.com/Homebrew/homebrew/master/README.md"
  version "1.0"
  depends_on "makedepend" => :build

  def install
    puts "Done."
  end
end
```
...or am I misunderstanding?

sjackman:
Thanks for the example, Matthew. That's perfect.

sjackman:
@lpar
> Error: pkg-config cannot be built with any available compilers.

Follow these instructions: https://github.com/Homebrew/linuxbrew/wiki/Standalone-Installation#create-symlinks-for-gcc

lpar:
Creating symlinks doesn't help. Same errors.

If you mean you want me to continue and install a whole new GCC and toolchain, that's a bit of a deal-killer.

sjackman:
@lpar You have two separate errors, Matthew. The solution to "Error: pkg-config cannot be built with any available compilers." is to create a symlink named `gcc-4.9` in your path pointing to `/usr/bin/gcc`.

The error "pkg-config not present or broken" we're troubleshooting here, but once you have the compiler symlink should be fixed with `brew install pkg-config`.

lpar:
Doesn't seem to be.
```
% brew install pkg-config
Error: pkg-config cannot be built with any available compilers.
To install this formula, you may need to:
  brew install gcc
% gcc-4.9 --version
gcc-4.9 (GCC) 4.9.2 20150212 (Red Hat 4.9.2-6)
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

jacknagel:
I can't reproduce this using upstream Homebrew. I don't have a linux VM or anything to test on, but if someone can give me access to such a system I'd be happy to investigate further.

sjackman:
@jacknagel If you have access to docker you can `docker pull sjackman/linuxbrew`

sjackman:
@lpar Hmm, that's frustrating. Can you report `which -a gcc-4.9` and `brew --prefix`?

If you're willing to try a `brew install gcc` which will install GCC 5.1, it will likely resolve your issue.

sjackman:
Also try `brew tap homebrew/versions`

jacknagel:
On a fresh Ubuntu install, I see this:

```
$ brew install -v bug264.rb 
==> Installing dependencies for bug264: pkg-config, makedepend
==> Installing bug264 dependency: pkg-config
==> Downloading http://pkgconfig.freedesktop.org/releases/pkg-config-0.28.tar.gz
```

lpar:
OK...

    % which -a gcc-4.9
    /usr/local/bin/gcc-4.9
    % ls -la /usr/local/bin/gcc-4.9 
    lrwxrwxrwx. 1 root root 8 2015-06-04 12:15 /usr/local/bin/gcc-4.9 -> /bin/gcc
    % ls -la /bin/gcc
    -rwxr-xr-x. 2 root root 858544 2015-02-12 05:12 /bin/gcc
    % ls -la /usr/bin/gcc
    -rwxr-xr-x. 2 root root 858544 2015-02-12 05:12 /usr/bin/gcc
 
I made the links manually. Obviously `/usr/local/bin` is on my path. `/usr/bin/gcc` and `/bin/gcc` are identical according to `sha256sum`.

I'm using Fedora 21 if that helps. Also, it occurs to me that I'm running `zsh`, which might not be the case for anyone else. Ruby version is 2.2.2p95, installed via `rbenv` and `ruby-build`.

`brew tap homebrew/versions` doesn't seem to help anything.

lpar:
OK, since it's a Friday I did some debugging, and I've found out why it's not detecting any working compilers.

In `Library/Homebrew/os/mac.rb`, `OS::Mac#locate`, it attempts to find GCC using `File.executable` on line 28 -- but it only looks in `/usr/bin`, it doesn't check any other path. If I move my symlinks to `/usr/bin` rather than `/usr/local/bin` then gcc-4.9 is found and I can `brew install pkg-config`.

sjackman:
@lpar It checks not only `/usr/bin/` but [two lines down](https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/os/mac.rb#L30) also `$(brew --prefix)/bin`. You should put your symlinks in there.

sjackman:
@jacknagel On Linuxbrew, the installation of bottles is disabled unless the user has the `glibc` and `patchelf` formulas installed, because those two formula are required to be able to use Linuxbrew bottles (see [here for code](https://github.com/Homebrew/linuxbrew/blob/37f6ab9c5da46c9e97a1e298a0094f537ca71362/Library/Homebrew/formula_installer.rb#L71)). Perhaps that logic is somehow related to this issue. I'll do some troubleshooting, see if I can reproduce the issue, and get back to you.

lpar:
Yes, it does also check `~/.linuxbrew/bin`. However, I think it ought to check the path and not require that the user create symlinks in order to use the distro's standard compiler.

sjackman:
Homebrew/Linuxbrew ignores most of the users' environment variables, including `PATH` by design to ensure reproducible builds.

> and not require that the user create symlinks in order to use the distro's standard compiler

I agree with this. I simply haven't got around to fixing the issue. Feel free to submit a pull request.

sjackman:
Note that nice distros like Debian and Ubuntu that include versioned compiler names like `/usr/bin/gcc-4.9` work just fine.

sjackman:
I've finally resolved this issue. I had modified `pour_bottle?` to check that `glibc` and `patchelf` are installed, but I had not modified `install_bottle_for?`. I mistakenly thought that `install_bottle_for?` calls `FormulaInstaller::pour_bottle?`, but it's actually calling `Formula::pour_bottle?`, which always returns `true`.

PR #266, opened 2015-01-28
superenv support in linuxbrew

veprbl:
This is a draft of a change that is to provide superenv support for linuxbrew. This is mostly to start discussion. I don't consider this to be a final solution.


Library/ENV/4.3/cc
```diff
@@ -1,6 +1,7 @@
-#!/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby -W0
+#!/usr/bin/ruby -W0
 
-$:.unshift Dir["/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/lib/ruby/{1.8,2.0.0}"].first
+lpath = Dir["/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/lib/ruby/{1.8,2.0.0}"].first
+$:.unshift lpath unless lpath.nil?
```
saketkc: This would still not work on linux, right?

veprbl: Yes. Without this change it would try to prepend nil to $: which in turn broke ```require``` directives.

veprbl: I meant: this part does work on linux for me.

sjackman:
Wow, thanks, Dmitry. I've been meaning to look into this. It's going to take me some time to review this, and I haven't had a whole lot of time for major Linuxbrew work recently. I am very interested in this patch though. Thanks for your work.

Library/ENV/4.3/cc
```diff
@@ -1,6 +1,7 @@
-#!/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby -W0
+#!/usr/bin/ruby -W0
```
sjackman: I usually use Homebrew's ruby. Let's use `#!/usr/bin/env ruby` instead.

Library/ENV/4.3/xcrun
```diff
@@ -3,16 +3,18 @@
 # work at all (e.g. CLT-only in the Xcode 4.3 era). This script emulates
 # it and attempts to avoid these issues.
 
-# Some build tools set DEVELOPER_DIR, so discard it
-unset DEVELOPER_DIR
+if [ "$(uname -s)" == "Darwin" ]; then
```
sjackman: `-s` is the default. Please remove it. For short commands like this, I prefer backticks.

sjackman:
The patch is surprisingly short. I'm excited that it may be so easy. I'll give it a test next week.

Library/ENV/4.3/cc
veprbl: I'm not sure what was the original intent of using ruby from framework. It would be great to hear what upstream thinks about this.

If framework's ruby is a necessity, we could have a ruby script in the superenv to call a correct ruby version depending on the system.

veprbl:
@sjackman Thanks for your interest. Unfortunately the change is not fully working yet. There is some problem that for many packages causes compilation fail at the linking stage.

sjackman:
Does `hello` build and install?

veprbl:
Two last commits solved all problems that I had so far.

sjackman:
Great! That's good news. I'll hopefully find the time to merge this next week.

veprbl:
What is linuxbrew's upstreaming policy?

sjackman:
Do we need the `ENV/linux` directory of symlinks, and can we just use the `4.3` directory instead?

veprbl:
My idea for separate directory in ENV was that we might want to have some utils behave differently or not have some utils at all (xcrun, mig, i686-apple-*).

sjackman:
That's seem quite likely, but I prefer not to anticipate trouble and skip the symlinks for now.

sjackman:
I'm getting ready to split the `linuxbrew` repository into `brew` and `homebrew-core`, as has been done for upstream `homebrew`. I'm not going to get around to merging this PR before then. So I'm finally closing this PR as `wontfix`, without however any prejudice against implementing `superenv` in the future. I simply never got around to committing the time necessary to test it.

sjackman:
Superenv has been added to Linuxbrew! https://github.com/Linuxbrew/brew/pull/263

veprbl:
Cool!

PR #269, opened 2015-01-31
cmake: fix python dependency

veprbl:
Systemwide python might not have setuptools on linux, so we better brew it.

### DIFF
--- a/Library/Formula/cmake.rb
+++ b/Library/Formula/cmake.rb
@@ -32,7 +32,7 @@ class Cmake < Formula
   end
 
   option "without-docs", "Don't build man pages"
-  depends_on :python => :build if OS.mac? && MacOS.version <= :snow_leopard && build.with?("docs")
+  depends_on :python => :build if ((OS.mac? && MacOS.version <= :snow_leopard) || OS.linux?) && build.with?("docs")
 
   depends_on "xz" # For LZMA
 


sjackman:
I'm thinking that it would make sense in [python_dependency.rb](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/python_dependency.rb#L8-L15) to check whether `setuptools` is installed. Could you try that instead? If it works, you can send the PR upstream to Homebrew rather than to Linuxbrew.

veprbl:
Then what about formulas that want python, but don't need setuptools?

sjackman:
Good question. I believe on Mac it's assumed that setuptools is installed—I'll have to confirm that. With Linuxbrew the general policy is to make the same assumptions that Mac Homebrew makes. Another option is to add a `--with-brewed-python` option that uses the Homebrew python rather than the system Python.

sjackman:
I don't want to force installing a brewed Python on a Linux system that already has Python installed.

sjackman:
Did you get an error message that `setuptools` was missing, and if so, can you post it?

veprbl:
```
==> python -c import setuptools... --no-user-cfg install --prefix=/home/tmp/cmake-KBdMXW/cmake-3.1.1/sphinx --single-version-externally-managed --record=installed.txt
--record=installed.txt

Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named setuptools
```

veprbl:
The fact is that ```--with-docs``` does depend on python. This should mean unconditional ```depends_on :python => :build```.

See also:
* 464418c588ba0ca4506f2ca4b92073c17b16e359
* Homebrew/homebrew#33356
* And the monography: http://wiki.kumokasumi.org/doku.php?id=homebrew_python_policy_draft

veprbl:
And that is understandable. Upstream can and will rely on uniformity OSX provides. On linux I would prefer to assume nothing because there are many different systems outdated each in their own way.

sjackman:
Brad (@chapmanb) contributed the commit ab148e4f44e5192abce317362e66bbe5d673872a that disabled the Python dependency for cmake. You and he should discuss.

sjackman:
The current policy of a standard Linuxbrew installation is that everything provided by the most recent version of Mac OS is provided by the host distribution of Linux, including Python and setuptools. Any formula can assume that this dependencies are already installed and do not require an explicit depends_on, for compatibility with upstream Homebrew.

To provide a better user experience, I would like to modify [python_dependency.rb](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/python_dependency.rb#L8-L15) to check that Python is in fact installed and provides all the same packages as Mac OS, and if not, install the `python` formula.

chapmanb:
Dmitry and Shaun;
My motivation for not wanting to pull in Python with cmake is that it is only required for docs and is a heavy requirement. Python brings in a lot of system level dependencies through brew. Since cmake is a requirement for a number of tools, you end up needing to bootstrap a large brew install before being able to install anything. For anyone trying to to use brew as a supplement to existing Linux package managers, this creates a big overhead.

An alternative solution is to require a flag to build the docs, instead of vice versa, so Python only gets pulled in and docs built if explicitly specified.

sjackman:
@chapmanb `cmake` does have a `--without-docs` option.

chapmanb:
Shaun;
Sorry to be confusing. I meant if the default was to not build docs that would be helpful. Things work okay if you explicitly install cmake with the `--without-docs` flag but do not if cmake gets built as a dependency for another package. As long as everything is explicit about building it up front we are good, but it's easy to trip into the Python build issues.

sjackman:
If the host OS provides Python, it probably also has `setuptools` installed. I'd like to check if that's the case and then use the host Python rather than installing Linuxbrew Python. I'm not keen on changing the default `--with-docs` to be different than Homebrew or having the default behaviour depend on whether Python is installed.

sjackman:
> Systemwide python might not have setuptools on linux, so we better brew it.

I'm going to assume that if the system has Python installed, it also has `setuptools` installed. I realize this may not always be true. Is there a common system where this is untrue?

dunn:
Just as a couple data points, neither xubuntu 14.04.2 nor the DigitalOcean distribution of Ubuntu 14.04 have setuptools.

sjackman:
That's good to know. Are [these instructions to install setuptools](https://www.digitalocean.com/community/tutorials/common-python-tools-using-virtualenv-installing-with-pip-and-managing-packages#installing-pip) useful?
`curl https://bitbucket.org/pypa/setuptools/raw/bootstrap/ez_setup.py | python -`

dunn:
Yeah, that would have worked, but I just went ahead and installed brew python.

sjackman:
The better solution, in my opinion. =)

PR #270, opened 2015-02-02
aria2: remove dependency on appletls

HorayNarea:
Linux doesn't have appletls, using OpenSSL instead

### DIFF
--- a/Library/Formula/aria2.rb
+++ b/Library/Formula/aria2.rb
@@ -24,6 +24,7 @@ class Aria2 < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "openssl"
 
   needs :cxx11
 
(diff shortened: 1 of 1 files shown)


Library/Formula/aria2.rb
```diff
@@ -24,6 +24,7 @@ class Aria2 < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "openssl"
```
sjackman: Please add `depends_on "openssl" unless OS.mac?`

Library/Formula/aria2.rb
```diff
@@ -31,8 +32,8 @@ def install
     args = %W[
       --disable-dependency-tracking
       --prefix=#{prefix}
-      --with-appletls
-      --without-openssl
+      --without-appletls
+      --with-openssl
```
sjackman: Use `if OS.mac?` and `if OS.linux?` as appropriate. Linuxbrew supports both Mac and Linux. (Homebrew supports only Mac).

sjackman:
Thanks for the patch.

PR #272, opened 2015-02-04
brew-config: fix for linuxbrew

xu-cheng:


sjackman:
Thanks for the patch.

PR #275, opened 2015-02-06
envv: New formula

bukzor:
test and audit pass

This project hasn't changed in a few years, but it's been around since 1994, and the maintainers are responsive whenever I need something.


sjackman:
Assuming this formula would also run on Mac it should be submitted to Homebrew/homebrew rather than Homebrew/linuxbrew.

bukzor:
https://github.com/Homebrew/homebrew/pull/36629

PR #277, opened 2015-02-07
fix library linking for python in linux

nemith:
The default sitecustomize points to OS specific paths.  This allows for
an option for a path to the linux library.

### DIFF
--- a/Library/Formula/python.rb
+++ b/Library/Formula/python.rb
@@ -60,6 +60,11 @@ class Python < Formula
     lib_cellar/"site-packages"
   end
 
+  def opt_libpython
+    opt_prefix/ (OS.mac? ? "Frameworks/Python.framework/Versions/2.7/Python" :
+                 "lib/libpython2.7.so")
+  end
+
   # The HOMEBREW_PREFIX location of site-packages.
   def site_packages
     HOMEBREW_PREFIX/"lib/python2.7/site-packages"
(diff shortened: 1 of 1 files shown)


sjackman:
Thanks for the patch. Can you give me a command that would fail otherwise without this patch, so that I can test that this patch resolves the issue?

nemith:
You need to compile something that links the python libraries.  I believe for me it was installing weechat with python support.

sjackman:
Compiling `libgcrypt` on my machine gives errors, so I'm not able to test installing `weechat`. Sorry. Are you able to give me another suggestion?

```
/bin/sh: line 5: 29023 Segmentation fault      GCRYPT_IN_REGRESSION_TEST=1 ${dir}$tst
FAIL: basic
…
/bin/sh: line 5: 29154 Segmentation fault      GCRYPT_IN_REGRESSION_TEST=1 ${dir}$tst
FAIL: benchmark
…
FAIL: bench-slope
SKIP: hashtest-256g
======================================
3 of 24 tests failed
(1 test was not run)
Please report to http://bugs.gnupg.org
======================================
Makefile:714: recipe for target 'check-TESTS' failed
make[2]: *** [check-TESTS] Error 1
```

sjackman:
I disabled the `make check` of `libgcrypt` and replicated your issue installing `weechat`. The `sitecustomize.py` of Homebrew seems to address Mac-specific issues. I disabled it by removing `sitecustomize.py`. Does that solution work for you?
```sh
rm -f `brew --prefix`/python2.7/site-packages/sitecustomize.pyc 
mv `brew --prefix`/lib/python2.7/site-packages/sitecustomize.py `brew --prefix`/lib/python2.7/site-packages/sitecustomize.py.orig
```

sjackman:
See https://github.com/Homebrew/linuxbrew/blob/master/Library/Formula/python.rb#L167
Possible fix
```diff
-     (site_packages/"sitecustomize.py").atomic_write(sitecustomize)
+    (site_packages/"sitecustomize.py").atomic_write(sitecustomize) if OS.mac?
```

nemith:
this works.  Thanks!

PR #280, opened 2015-02-08
icecast: depends on libxslt and curl

iggyvolz:
See Homebrew/homebrew#36636.

In trying to install this, I got numerous errors. Fixed by adding a depends_on libxslt and depends_on curl.

These are installed by default on macs, but they must be installed on a non-Mac platform i.e. an Ubuntu with linuxbrew.

### DIFF
--- a/Library/Formula/icecast.rb
+++ b/Library/Formula/icecast.rb
@@ -17,6 +17,8 @@ class Icecast < Formula
   depends_on 'speex'  => :optional
   depends_on 'openssl'
   depends_on 'libvorbis'
+  depends_on 'libxslt' unless OS.mac?
+  depends_on 'curl' unless OS.mac?
 
   def install
     system "./configure", "--disable-dependency-tracking",


sjackman:
Linuxbrew currently assumes that all software provided by Mac OS is already installed using the host package management system.

iggyvolz:
Is this the "sudo apt-get install build-essential curl..." line on the README?  I thought I did that but I may have forgotten on that particular server.

Could one make these :optional?  The Ubuntu system usually gets out of date and I prefer to use linuxbrew.

sjackman:
Yes, I think I'm going to start including these dependencies as
`depends_on "curl" => :recommended unless OS.mac?`

sjackman:
Could you revise this PR?

Library/Formula/icecast.rb
```diff
@@ -1,5 +1,3 @@
-require 'formula'
-
```
sjackman: To make merging with Homebrew easier, patches to Linuxbrew should not include unnecessary changes. Changes such as these can be submitted upstream to Homebrew, but they do not accept style-only changes.

iggyvolz: Whoops - so used to doing that now that it's a matter of instinct.  I'll take that out.

iggyvolz:
How do you want the commit message to be styled for this?  Is "Update icecast dependencies" good or should I change it to the title?

sjackman:
If it fits in 50 characters, then `icecast 2.4.1: Requires libxslt and curl on Linuxbrew`
otherwise `iceast 2.4.1: Add dependencies for Linuxbrew` and list the dependencies in the message.

sjackman:
Thanks for you contribution, Iggy.

iggyvolz:
No problem!  Any time!

PR #282, opened 2015-02-09
intltool: depends on xml::parser for linuxbrew

saketkc:
See travis build failure for https://github.com/Homebrew/homebrew-science/pull/1806
This won't help either, but would at least would fail explicitly pre-brewing.


sjackman:
I think I ran into this issue myself as well.

sjackman:
I guess we need to add `brew install cpanm && cpanm XML::Parser` to the base installation instructions, though I'm not super keen on that.

PR #291, opened 2015-02-23
Task Warrior: bumped versions and cleanup

JensErat:
- Bumped version to 2.4.1
- Changed HEAD to 2.4.2 branch (where development goes on)
- Removed OS X bottles


JensErat:
Seems I have to put some additional effort in this. Worked for me, but there seems to be some (new) dependency missing. Will open a new pull request when resolved.

sjackman:
New versions should be submitted to Homebrew/homebrew rather than Homebrew/linuxbrew. Homebrew is merged into Linuxbrew once per week.

PR #294, opened 2015-02-25
task: depends on libuuid

JensErat:


JensErat:
There is another problem, which I'm not sure about what to do:

> cc1plus: error: unrecognized command line option ‘-std=c++11’

I have no issues on my system. Is it possible the rather old [GCC 4.6 does not support C++11](https://gcc.gnu.org/wiki/C11Status)? If so, how could this be resolved (for Travis)?

sjackman:
> Is it possible the rather old GCC 4.6 does not support C++11?

Yes. Install GCC 4.9 using Homebrew: `brew install gcc`

JensErat:
So should task reference GCC 4.9 as `on_build` dependency? As far as I know the Formula Cookbook (nor scrolling through it again) we cannot enforce something like "at least GCC 4.7, otherwise install GCC from homebrew), or is there a better alternative (for users with adequate version of GCC)?

sjackman:
To require a compiler that supports C++11 use:
```ruby
needs :cxx11
```

sjackman:
The libuuid dependency won't be accepted in Homebrew/homebrew, because libuuid is provided by Mac OS, so I'll merge this PR into Linuxbrew. The C++11 requirement should be submitted to Homebrew/homebrew.

sjackman:
Thanks for your contribution, Jens.

JensErat:
I will propose the C++11 patch upstream, thank you for the hint. Is it possible to somehow trigger travis builds _without_ pull requests? I guess we have very similar problems for libuuid, but they're rather hard to replay on my local machine.

Or some docker container with a "travis-like" setup. I see you have added some on the [index](https://hub.docker.com/u/sjackman/), are any of these comparable to Travis?

sjackman:
You can create a pull request, and every push to that PR will trigger a Travis build. I'm fine with that. Just make it clear in the PR title what you're up to. You can also fork Linuxbrew and run Travis in your own fork. I'm not familiar with a Travis Docker image, but I'm almost certain that one should exist.

PR #297, opened 2015-03-01
sip: Fix for linuxbrew

saketkc:


saketkc:
Avoiding the single line change that would have been easier, since the empty space generated breaks to `configure.py` script.

Library/Formula/sip.rb
```diff
@@ -30,12 +30,15 @@ def install
 
     Language::Python.each_python(build) do |python, version|
       # Note the binary `sip` is the same for python 2.x and 3.x
-      system python, "configure.py",
-                     "--deployment-target=#{MacOS.version}",
```
sjackman: Does  the following work? I'd prefer the smaller patch if it does.
```diff
- "--deployment-target=#{MacOS.version}",
+ ("--deployment-target=#{MacOS.version}" if OS.mac?),
````

saketkc: That's what I tried. It still leads to an empty space, and `configure.py` recognises it as an extra option.

sjackman:
Thanks for your contribution, Saket.

You can use `compact` to remove the `nil` entry from the vector. See here: https://github.com/Homebrew/linuxbrew/commit/0c2c2cfc7c0c85695d80debfec5220d2de527cd6

PR #298, opened 2015-03-02
a2ps 4.14: Requires gperf on Linuxbrew

iggyvolz:
Per #219 and new policy in #280, require homebrew/dupes/gperf for Linuxbrew.  (I'm just going through formula to see if they install properly without using apt-get dependencies using throwaway cloud VM's).  There may be a better way to go about it, maybe a homebrw-linuxbrew-dupes?

I'm getting a possibly unrelated error while installing.  That may be because gperf itself isn't perfected for linux.  https://gist.github.com/anonymous/f0be8f2ba8ea64fbbff8.  It might go away in tests.

### DIFF
--- a/Library/Formula/a2ps.rb
+++ b/Library/Formula/a2ps.rb
@@ -4,6 +4,8 @@ class A2ps < Formula
   mirror "http://ftp.gnu.org/gnu/a2ps/a2ps-4.14.tar.gz"
   sha1 "365abbbe4b7128bf70dad16d06e23c5701874852"
 
+  depends_on "homebrew/dupes/gperf" unless OS.mac?
+
   bottle do
     sha1 "c33f22a088a0b1ed22efff0165722e87495a4bd0" => :yosemite
     sha1 "7ae09c9835ebb1913b97cf5f06bab42a0d1f33a6" => :mavericks


sjackman:
https://gist.github.com/anonymous/f0be8f2ba8ea64fbbff8#file-02-make-L648-L655
```
/usr/bin/gcc-4.8 -Os -w -pipe -march=core2 -Wl,-rpath -Wl,/home/ubuntu/workspace/linuxbrew/lib -o sample main.o  -L/home/ubuntu/workspace/linuxbrew/lib -lm ../../lib/.libs/liba2ps.a  
../../lib/.libs/liba2ps.a(metaseq.o): In function `grow_user_string_obstack':
metaseq.c:(.text+0x7011): warning: the use of `tempnam' is dangerous, better use `mkstemp'
../../lib/.libs/liba2ps.a(psgen.o): In function `virtual_begin':
psgen.c:(.text+0x7ad): undefined reference to `atan2'
collect2: error: ld returned 1 exit status
make[2]: *** [sample] Error 1
make[2]: Leaving directory `/tmp/a2ps-Rpiuox/a2ps-4.14/contrib/sample'
```

sjackman:
For `undefined reference to `atan2'`, `-lm ../../lib/.libs/liba2ps.a` should perhaps be `../../lib/.libs/liba2ps.a -lm`.

sjackman:
Thanks again for your contribution, Iggy.

PR #300, opened 2015-03-03
a52dec 0.7.4: Requires CPPFLAGS=-fPIC

iggyvolz:
a52dec 0.7.4 requires CPPFLACS=-fPIC to compile on linux (log of failure in current state: https://gist.github.com/anonymous/7dddfe49dafd8070448d).

I don't know exactly how this should be merged.  Since it is a system command, I can't exactly append the CPPFLAGS=-fPIC on a new line.  My ruby is crap so I did the best I could (an unless OS.mac and unless not OS.mac for the old behavior) - it doesn't look good and will likely be a headache for merge conflicts in Homebrew but it's a start.

### DIFF
--- a/Library/Formula/a52dec.rb
+++ b/Library/Formula/a52dec.rb
@@ -11,6 +11,7 @@ class A52dec < Formula
                           "--prefix=#{prefix}",
                           "--enable-shared",
                           "--mandir=#{man}"
-    system "make install"
+    system "make install CPPFLAGS=-fPIC" unless OS.mac?
+    system "make install" unless not OS.mac?
   end
 end


sjackman:
That looks like an upstream bug to me. Please open a bug upstream. It is odd though. They're using libtool, which should handle adding `-fPIC`.

```
/gsc/btl/linuxbrew/bin/gcc-4.9 -DHAVE_CONFIG_H -I. -I. -I../include -I../include -I../include -isystem/gsc/btl/linuxbrew/include -Os -w -pipe -march=core2 -fomit-frame-pointer -c imdct.c -o .libs/imdct.lo
…
/gsc/btl/linuxbrew/bin/gcc-4.9 -shared  bitstream.lo imdct.lo bit_allocate.lo parse.lo downmix.lo  -L/gsc/btl/linuxbrew/lib -lm  -Wl,--dynamic-linker=/gsc/btl/linuxbrew/opt/glibc/lib/ld-linux-x86-64.so.2 -Wl,-rpath -Wl,/gsc/btl/linuxbrew/lib -Wl,-soname -Wl,liba52.so.0 -o .libs/liba52.so.0.0.0
/gsc/btl/linuxbrew/bin/ld: imdct.lo: relocation R_X86_64_32 against `.bss' can not be used when making a shared object; recompile with -fPIC
imdct.lo: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
Makefile:196: recipe for target 'liba52.la' failed
make[1]: *** [liba52.la] Error 1
```

sjackman:
Does a52dec compile correctly outside of Linuxbrew?

sjackman:
Try
```ruby
args = []
args << "CPPFLAGS=-fPIC" unless OS.mac?
system "make", "install", *args
```

sjackman:
Even better is
```ruby
    ENV.append_to_cflags "-fPIC" unless OS.mac?
```

PR #310, opened 2015-03-13
Improve support for RPM-based distributions

tseemann:


sjackman:
@tseemann The group "Additional Development" installs a fair number of packages, most of which aren't strictly necessary to Linuxbrew (I don't think). Was there a particular package that was missing and caused you a dependency issue?

```
   GConf2-devel
   alsa-lib-devel
   audit-libs-devel
   binutils-devel
   boost-devel
   bzip2-devel
   cyrus-sasl-devel
   dbus-glib-devel
   elfutils-devel
   elfutils-libelf-devel
   evolution-data-server-devel
   gnome-desktop-devel
   gnome-keyring-devel
   gstreamer-devel
   gvfs-devel
   hmaccalc
   java-1.6.0-openjdk-devel
   java-1.7.0-openjdk-devel
   libacl-devel
   libattr-devel
   libcanberra-devel
   libcap-ng-devel
   libcgroup-devel
   libcurl-devel
   libgnome-devel
   libnotify-devel
   librsvg2-devel
   libselinux-devel
   libsoup-devel
   mysql-devel
   perl-ExtUtils-MakeMaker
   perl-devel
   polkit-devel
   postgresql-devel
   pulseaudio-libs-devel
   pygtk2-devel
   python-devel
   readline-devel
   sqlite-devel
   tbb-devel
```

tseemann:
The problem was that the system uses RHEL and the 'Satellite' management system, which means only the head sysadmin can add channels.  He has now installed that channel.

Is there any way we can put hints in recipes for this? 

     depends_on 'gd' => :brew, 'gd-devel' => :rpm, 'libgd-dev' => :apt 

Like the way we do for Perl? 

      depends_on 'Foo::Bar' => :perl 

Or in some kind of clause/method which won't annoy the Mac OS X team?

Or some master list in LinuxBrew which gets triggered?

sjackman:
What is the ultimate goal? Is it to avoid installing the Homebrew package (like `gd`) if that dependency is satisfied by the native package manager?

tseemann:
I guess so. The Mac Brew team have the easier situation of generally knowing exactly what -dev packages Mac OS X provide. I was hoping to make Linux Brew closer to that ideal. But it is pointless as RPM/APT need sudo anyway, and Brew is all about non-sudo. 

I normally have full control over my hardware but I'm in a new situation where I don't always. And when lots of -dev packages are not installed, it reveals a lot of "fluke" formulae that are actually using system packages without realising it. This seems to violate some of the goals of Brew, as you have external package dependencies which you can't control and may get upgraded and break your Brew install.

Writing good formulae is hard. A "blank" Linux system for Jenkins would reveal most of these errors. 

But ultimately it is probably not worth the effort. Linuxbrew is working well, and forcing me to write Mac compatible formulae helps the greater community. I do have a Mac laptop but it's a glorified dumb terminal :-)

sjackman:
That standard Linuxbrew installation uses both host-provided packages as well Linuxbrew packages. It assumes that the host provides all the same packages that are provided by Mac OS.

The standalone Linuxbrew installation uses only Linuxbrew packages. See the [Standalone Installation instructions](https://github.com/Homebrew/linuxbrew/wiki/Standalone-Installation).

> Writing good formulae is hard. A "blank" Linux system for Jenkins would reveal most of these errors.

Linuxbrew uses [Docker](https://hub.docker.com/u/sjackman/) and [Travis](https://travis-ci.org/Homebrew/linuxbrew) for testing.

PR #318, opened 2015-03-19
Openssl 1.0.2a - Security vulnerability merged from Homebrew

iggyvolz:
I cherry-picked the appropriate commits from openssl.rb.  Not sure if this is the correct method to do it but it really needs to get patched ASAP (notes: http://openssl.org/news/secadv_20150319.txt)

### DIFF
--- a/Library/Formula/openssl.rb
+++ b/Library/Formula/openssl.rb
@@ -8,9 +8,9 @@ class Openssl < Formula
   version "1.0.2a-1"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Mikaela:
I would say that the correct way is to open issue requesting quick merge from Homebrew as there is security vulnerability fixed there.

sjackman:
Merging Homebrew is the way to go. I'll do so today.

PR #322, opened 2015-03-23
boost: Add option to build boost with brewed bzip2.

yesimon:
Not sure if this is best place for it, but I was guessing linuxbrew because bzip2 headers and libs should typically be installed on OS X, but might not be on linux.

### DIFF
--- a/Library/Formula/boost.rb
+++ b/Library/Formula/boost.rb
@@ -96,6 +96,12 @@ class Boost < Formula
             "--user-config=user-config.jam",
             "install"]
 
+    if OS.linux?
+      args += ["include=#{HOMEBREW_PREFIX}/include",
+               "cflags=-L#{HOMEBREW_PREFIX}/lib",
+               "cxxflags=-L#{HOMEBREW_PREFIX}/lib"]
+    end
+
     if build.with? "single"
       args << "threading=multi,single"
     else


sjackman:
Hi, Simon. Were you getting an error message without this patch, and if so, what was it?

Have you seen the [Standalone Installation instructions](https://github.com/Homebrew/linuxbrew/wiki/Standalone-Installation) for Linuxbrew, and is that an option for you?

yesimon:
Something like couldn't find header bzlib.h while running ./b2 during boost installation. I think it has something to do with the boost.build build system which I don't really understand, and they even have a page documenting installing without system bzlib.h http://www.boost.org/doc/libs/1_57_0/libs/iostreams/doc/installation.html

The standalone installation looks like it could work but seems a bit heavy handed? Is that something you recommend for linuxbrew users without sudo to avoid having too many patches on specific formulae?

sjackman:
Do you still get that error if the `bzip2` formula is installed? If not, I'll add to the boost formula
```ruby
depends_on "bzip2" unless OS.mac?
```

sjackman:
> The standalone installation looks like it could work but seems a bit heavy handed? Is that something you recommend for linuxbrew users without sudo to avoid having too many patches on specific formulae?

I like standalone, because it helps clear up a lot of issues that stem from confusion about whether a given package is using the system library or the Homebrew library, because it only uses Homebrew libraries. It depends on your use case though. If you just want to install a few Linuxbrew packages, I would suggest the standard install simply for the lower barrier to entry. If you're planning to use tens or hundreds of Linuxbrew packages, I would use the standalone installation.

yesimon:
Yes I still get the error with ```depends_on "bzip2"``` or with bzlib2 installed. I could be totally wrong, but I think the issue is that Boost.Build is a different build system that doesn't really listen to things like ```CPPFLAGS```, which I believe is what brew uses to compile in the standard environment, hence the special env variables ```BZIP2_...```.

Plus I think it's right to give people the option whether to use the system bzip2 or a brewed bzip2? Instead of always using brewed bzip2 if installed on linux.

sjackman:
Can you submit this patch upstream to Homebrew/homebrew?

yesimon:
This got bounced back from Homebrew/homebrew#39018

sjackman:
Can you please post the gist: `brew gist-logs boost`
Which compiler are you using? You may want to try the brewed GCC 4.9 compiler `brew install gcc`

yesimon:
This is normal boost formula
https://gist.github.com/yesimon/8c8b89c2bfe85c003d43
(search for "bzlib.h" on 02.b2)

This is with the `--with-brewed-bzip2` change
https://gist.github.com/yesimon/5a5bceb77e2d44cc51d1

sjackman:
https://gist.githubusercontent.com/yesimon/8c8b89c2bfe85c003d43/raw/02.b2
```
gcc.compile.c++ bin.v2/libs/iostreams/build/gcc-4.9.2/release/threading-multi/bzip2.o

    "/idi/sabeti-scratch/yesimon/linuxbrew/bin/g++-4.9"  -ftemplate-depth-128 -O3 -finline-functions -Wno-inline -Wall -pthread -fPIC  -DBOOST_ALL_NO_LIB=1 -DBOOST_IOSTREAMS_DYN_LINK=1 -DBOOST_IOSTREAMS_USE_DEPRECATED -DNDEBUG  -I"." -c -o "bin.v2/libs/iostreams/build/gcc-4.9.2/release/threading-multi/bzip2.o" "libs/iostreams/src/bzip2.cpp"

libs/iostreams/src/bzip2.cpp:20:56: fatal error: bzlib.h: No such file or directory
 #include "bzlib.h"  // Julian Seward's "bzip.h" header.
                                                        ^
compilation terminated.
...failed gcc.compile.c++ bin.v2/libs/iostreams/build/gcc-4.9.2/release/threading-multi/bzip2.o...
```

sjackman:
Does Boost have an option to add `$HOMEBREW_PREFIX/include` and  `$HOMEBREW_PREFIX/lib` to the search directories for all libraries, not just bzip2?

yesimon:
Not that I know of. I'll ask the B2 mailing list but otherwise why would they have special vars like `BZIP2_INCLUDE` and `ZLIB_INCLUDE`?

sjackman:
To override the location of particular libraries when they're also found in the global search path.

yesimon:
I've changed it to include the linuxbrew include dir and library dirs by default. Tested and it works, but not sure if it will have any other side effects. PTAL

sjackman:
Looks fine to me. Patches for Linuxbrew should not change the behaviour on Mac. Please see [CONTRIBUTING.md](https://github.com/Homebrew/linuxbrew/blob/master/CONTRIBUTING.md)

Please use `args += […] unless OS.mac?`

yesimon:
Updated.

sjackman:
Merged. Thanks for your contribution, Simon.

PR #323, opened 2015-03-25
Make libedit optional for libslax/juise

nemith:
Juise and libslax will fall back to readline if libedit is not specified.

### DIFF
--- a/Library/Formula/juise.rb
+++ b/Library/Formula/juise.rb
@@ -26,11 +26,13 @@ class Juise < Formula
 
   def install
     system "sh ./bin/setup.sh" if build.head?
-    system "./configure", "--disable-dependency-tracking",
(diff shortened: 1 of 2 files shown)


sjackman:
What's the benefit of using readline over libedit? Are you getting an error when building these formulae? If so, please gist the error message. `brew gist-logs juise`

nemith:
I submitted maintain libslax/juise for homebrew OS X.  The reason the libedit peices were there is because readline is not included in OS X by default but libedit is (stop another dependency).   This is just reversing that logic for linux.

sjackman:
Is there any benefit to using readline instead of libedit on Linux? I'd be more inclined use the same dependency (libedit) on both systems.
```ruby
depends_on "homebrew/dupes/libedit" unless OS.mac?
```

PR #326, opened 2015-03-27
Upgrade jdk formula to jdk 1.8.0_40

aruizca:


Mikaela:
I think this should go to [Homebrew](https://github.com/Homebrew/Homebrew).

aruizca:
Thanks

sjackman:
On Mac, the JDK formula is in Homebrew-cask, but there is no cask support for Linuxbrew.

sjackman:
Thanks for your contribution, Angel.

PR #329, opened 2015-04-01
Fix various issues in python3 formula.

yesimon:
- ```Pathname``` doesn't like concating with nil.
- Fix some other issues with lib_cellar in python3.
- --enable-shared like in python2 formula

### DIFF
--- a/Library/Formula/python3.rb
+++ b/Library/Formula/python3.rb
@@ -50,10 +50,13 @@ class Python3 < Formula
   # X11.
   patch :DATA if build.with? "brewed-tk"
 
+  def lib_cellar
+    prefix / (OS.mac? ? "Frameworks/Python.framework/Versions/#{xy}" : "") /
+      "lib/python#{xy}"
+  end
+
   def site_packages_cellar
-    prefix/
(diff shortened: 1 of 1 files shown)


Library/Formula/python3.rb
```diff
@@ -50,10 +50,16 @@ class Python3 < Formula
...
   patch :DATA if build.with? "brewed-tk"
 
+  def lib_cellar
+    if OS.mac?
+      prefix/"Frameworks/Python.framework/Versions/2.7/lib/python2.7"
+    else
+      prefix/"lib/python#{xy}/site-packages"
+    end
+  end
+
```
sjackman: Please use identical code to `python.rb`. It'll make maintenance of the two formulae easier.
See https://github.com/Homebrew/linuxbrew/blob/master/Library/Formula/python.rb#L56-L59

sjackman:
Thanks, Simon.

Library/Formula/python3.rb
yesimon: Yeah whoops also have the 2.7 path here.

yesimon:
Updated.

sjackman:
Thanks!

PR #330, opened 2015-04-02
Determine sqlite include path for python on linux as well.

yesimon:
Python's setup.py extracts include/lib paths from -I, -L sysconfig vars when the os is darwin, but it even says in the comments it should work for unixy os as well! The point is that setup.py tries to build the sqlite3 extension looking for system sqlite3, which either may not exist, or is not what you want when linking to brew built sqlite3. This is why I think it shouldn't be a flag option like --with-brewed-sqlite.


sjackman:
Please submit this patch to `setup.py` upstream to Python.

Issue #333, opened 2015-04-02
binutils fails to install on xubuntu (14.04.2)


dunn:
There are a couple of open issues about binutils but those look like problems with upgrading existing installations; I just can't get it to install in the first place.

**[`gist-logs`](https://gist.github.com/1bcd0fe0d0e076e5f3e4)**

```
HOMEBREW_VERSION: 0.9.5
ORIGIN: https://github.com/Homebrew/linuxbrew.git
HEAD: 3ada3be8d941512cc649d530b75e03f0dd2cf719
Last commit: 21 hours ago
HOMEBREW_PREFIX: /home/alix/.linuxbrew
HOMEBREW_CELLAR: /home/alix/.linuxbrew/Cellar
CPU: dual-core 64-bit 6
OS X: 0-x86_64
Xcode: N/A
CLT: N/A
Clang: N/A
X11: N/A
System Ruby: 1.9.3-p484 => /usr/bin/ruby1.9.1
Perl: /usr/bin/perl
Python: /usr/bin/python => /usr/bin/python2.7
Ruby: /usr/bin/ruby => /usr/bin/ruby1.9.1
Java: N/A
```

dunn:
Spoke too soon: just had to `sudo apt-get install g++`.

Should I close this, or is this something that linuxbrew can/should check for?

sjackman:
Glad you got it working. See [Linuxbrew Dependencies](http://brew.sh/linuxbrew/#dependencies).

dunn:
Oh gotcha, thanks!  Would it be worth adding a note about dependencies to the end of the install script?  I can open a PR if you think it's a good idea.

sjackman:
I've started adding dependencies directly into the formulas to make this less of an issue. You'll definitely still need a C/C++ compiler, but the library dependencies should no longer be necessary. You could output a message saying either `sudo apt-get install build-essential` for Debian/Ubuntu or `sudo yum groupinstall 'Development Tools'` for Fedora/CentOS/Red Hat.

PR #336, opened 2015-04-04
install: list basic dependencies

dunn:
followup to #333


sjackman:
Thanks for this, Alex. Please remove the style-only changes from the patch. If they're from the upstream `go` branch, I'll merge them in a separate commit.

dunn:
Upstream is using a separate repository for the install script now, rather than the `go` branch, so I wasn't sure if it was still a priority to keep the branches in sync.

sjackman:
I'd rather keep the patch small. It makes it easier to review.

dunn:
Done.

sjackman:
Thanks, Alex.

PR #340, opened 2015-04-10
Allows the openssl formula to build when pkg-config is not present

tbetbetbe:
- pkg-config is not always installed
- makedepend has a build dependency on pkg-config, since openssl has build dependency on makedepend, it should also have a build dependency on pkg-config


sjackman:
Dependencies are resolved recursively. It shouldn't be necessary to explicitly list recursive dependencies.

PR #341, opened 2015-04-10
makedepend: Depends on pkg-config

tbetbetbe:
- pkg-config is not always installed
- makedepend has a build dependency on pkg-config, since openssl has build dependency on makedepend, it should also have a build dependency on pkg-config


sjackman:
It should not be necessary to explicitly list recursive dependencies, since dependencies are resolved recursively. Can you give an example to replicate this error?

PR #352, opened 2015-04-15
lua51: Fix for Linux

seanstrom:
This patches the Lua51 Formula to work Linux, specifically Ubuntu at the moment.
This was apparently patched before but is somehow was undone.
https://github.com/Homebrew/linuxbrew/issues/79

@sjackman


sjackman:
Thanks, Sean.

PR #363, opened 2015-04-20
Lua depends on readline, close #359

### LINKED ISSUE #359: Lua is missing a readline dependency
see https://github.com/Homebrew/homebrew/pull/37681

Could you please add
```
depends_on "readline" unless OS.mac?
``` 
I did not create a PR as I do not have `linuxbrew` forked and don't really modify it that much.

iggyvolz:


sjackman:
Thanks, Iggy.

PR #383, opened 2015-04-29
Cmake depends on curl (close #381)

### LINKED ISSUE #381: cmake depends on curl
See https://github.com/tseemann/homebrew-bioinformatics-linux/issues/1

iggyvolz:
close #381.

### DIFF
--- a/Library/Formula/cmake.rb
+++ b/Library/Formula/cmake.rb
@@ -15,6 +15,7 @@ class Cmake < Formula
   option "without-docs", "Don't build man pages"
 
   depends_on :python => :build if OS.mac? && MacOS.version <= :snow_leopard && build.with?("docs")
+  depends_on "curl" unless OS.mac?
 
   # The `with-qt` GUI option was removed due to circular dependencies if
   # CMake is built with Qt support and Qt is built with MySQL support as MySQL uses CMake.


Library/Formula/cmake.rb
```diff
@@ -15,6 +15,7 @@ class Cmake < Formula
   option "without-docs", "Don't build man pages"
 
   depends_on :python => :build if OS.mac? && MacOS.version <= :snow_leopard && build.with?("docs")
+  depends_on "curl"
```
sjackman: curl is provided by default on Mac OS, so this dependency should not apply to Mac OS. Linuxbrew aims to maintain compatibility with Homebrew and not change the behaviour of Homebrew when Linuxbrew is run on a Mac. This line should be
```ruby
depends_on "curl" unless OS.mac?
```

iggyvolz: I knew that, sorry!

sjackman:
To be consistent the commit message should start with `cmake:` rather than `Cmake`,  note the case  and colon.

sjackman:
Please also include "for Linuxbrew" in the first line of the commit message.

sjackman:
Thanks, Iggy.

PR #385, opened 2015-05-01
doxygen: Depends on flex and bison

saketkc:


Library/Formula/doxygen.rb
```diff
@@ -21,6 +21,8 @@ class Doxygen < Formula
   depends_on "graphviz" => :optional
   depends_on "qt" if build.with? "doxywizard"
   depends_on "llvm" => "with-clang" if build.with? "libclang"
+  depends_on "flex" if OS.linux?
+  depends_on "bison" if OS.linux?
```
sjackman: Please change these both from `if OS.linux?` to `unless OS.mac?`

sjackman:
Hi, Saket. Would you like to fix up this PR?

saketkc:
Fixed

sjackman:
Thanks, Saket.

PR #335, opened 2015-04-03
Fix rtmpdump and libbluray for Linux

pgkos:
rtmpdump doesn't build on Linux. I have added a fix.
I also have added libxml2 dependency to libbluray (only on Linux, to simplify setup).

PR #386, opened 2015-05-03
Fix rtmpdump for Linux

pgkos:
follow-up to #335


sjackman:
Thanks for your contribution.

PR #387, opened 2015-05-03
libbluray: Depends on libxml2

pgkos:
follow-up to #335


Library/Formula/libbluray.rb
```diff
@@ -22,10 +22,11 @@ class Libbluray < Formula
   depends_on "pkg-config" => :build
   depends_on "freetype" => :recommended
   depends_on "fontconfig"
+  depends_on "libxml2" if OS.linux?
```
sjackman: Please use `unless OS.mac?` rather than `if OS.linux?` for dependencies.

sjackman:
Thanks for your contribution, @pgkos.

PR #397, opened 2015-05-08
fontconfig: set POSIXLY_CORRECT when building

ntkme:


sjackman:
Hi, @ntkme Does this patch work around some error message that you're seeing?

ntkme:
@sjackman I just tested on another Linux distro, and it failed because of following:

```
grep: character class syntax is [[:space:]], not [:space:]
```

`ENV["POSIXLY_CORRECT"]` fixed it, so I updated the patch.

sjackman:
On which Linux distribution did you see that error?

ntkme:
CentOS 6 with brewed `grep`. System `grep` doesn't have this problem.

sjackman:
Can you please report the output of `brew gist-logs fontconfig`?

ntkme:
https://gist.github.com/anonymous/b1aa70bf6a0530737221

sjackman:
https://gist.github.com/anonymous/b1aa70bf6a0530737221#file-02-make-L316-L326
```
docbook2man local-fontconfig-user.sgml && \
rm -f manpage.*
grep: character class syntax is [[:space:]], not [:space:]
grep: character class syntax is [[:space:]], not [:space:]
jw: There is no frontend called "/docbook/utils-0.6.14/frontends/docbook".
Makefile:1310: recipe for target 'fonts-conf.5' failed
make[2]: *** [fonts-conf.5] Error 2
make[2]: Leaving directory '/tmp/fontconfig20150511-8139-poipx2/fontconfig-2.11.1/doc'
Makefile:1167: recipe for target 'install' failed
make[1]: *** [install] Error 2
```

sjackman:
Do you have `docbook2x` installed? Does installing `docbook2x` help?

```sh
which -a docbook2man
brew list --versions docbook2x
brew install docbook2x fontconfig
```

ntkme:
No, it doesn't help. It actually has nothing to do with docbook2man. As I said, the problem is on GNU `grep`.

Setting `POSIXLY_CORRECT` forces GNU `grep` behave under posix standard.

sjackman:
Based on [the error message quoted above](https://github.com/Homebrew/linuxbrew/pull/397#issuecomment-101035571) I was guessing that the incorrect usage of grep was in docbook2man. I couldn't find this incorrect usage of grep anywhere in fontconfig. This issue looks like an upstream bug to me. It'd be better to resolve it upstream if possible.

sjackman:
Please report
```sh
which -a docbook2man
docbook2man --version
```

ntkme:
``` sh
$ which -a docbook2man
/data/linuxbrew/bin/docbook2man
/usr/bin/docbook2man
$ docbook2man --version
/data/linuxbrew/bin/docbook2man (part of docbook2X 0.8.8)
$Revision: 1.12 $ $Date: 2006/04/14 17:29:04 $
<URL:http://docbook2x.sourceforge.net/>

Copyright (C) 2003-2004 Steve Cheng
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

sjackman:
Please report this bug upstream to fontconfig and post a link to the issue here. It would be better fixed upstream in fontconfig than hacked around in Linuxbrew. Thanks for identifying the workaround though.

sjackman:
It's possible that the bug's in a dependency of fontconfig, but maybe the fontconfig developers can help figure out where the true cause lies.

ntkme:
It has been fixed in upstream...

And I found a better workaround for this:

``` sh
brew unlink grep
brew install fontconfig
brew link grep
```

PR #402, opened 2015-05-12
Link to Linuxbrew issue tracker, not Homebrew

### LINKED ISSUE #401: Fix troubleshooting links so that Homebrew doesn't get spammed
The links mentioned [here](https://github.com/Homebrew/linuxbrew/blob/master/share/doc/homebrew/Troubleshooting.md#check-to-see-if-the-issue-has-been-reported) and [here](https://github.com/Homebrew/linuxbrew/blob/master/share/doc/homebrew/Troubleshooting.md#create-an-issue) point to the Homebrew issue tracker.  They should be changed to point to the Linuxbrew tracker.

This might explain, in part, the rash of Linuxbrew problems that get reported to Homebrew.

iggyvolz:
fix #401

### DIFF
--- a/share/doc/homebrew/Troubleshooting.md
+++ b/share/doc/homebrew/Troubleshooting.md
@@ -27,12 +27,12 @@ Thank you!
 * If `brew doctor` warns about unbrewed dylibs, be advised that they are **very likely to cause build problems**.
 
 ## Check to see if the issue has been reported
-* Browse open issues on the [issue tracker](https://github.com/Homebrew/homebrew/issues) to see if someone else has already reported the same problem.
+* Browse open issues on the [issue tracker](https://github.com/Homebrew/linuxbrew/issues) to see if someone else has already reported the same problem.
 * Make sure you check issues on the correct repository. If the formula that failed to build is part of a tap, like [homebrew-science](https://github.com/Homebrew/homebrew-science) or [homebrew-dupes](https://github.com/Homebrew/homebrew-dupes), check there instead.
 
 ## Create an issue
 
-1. [Create a new issue](https://github.com/Homebrew/homebrew/issues/new)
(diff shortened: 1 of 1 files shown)


iggyvolz:
I think I found a couple more instances of linking to homebrew in the docs, so I'll add these in.

iggyvolz:
Not sure if this instance should be replaced: https://github.com/Homebrew/linuxbrew/blob/5d1ef0f4a670cccb5835327bfbfede40cbb468eb/share/doc/homebrew/Formula-Cookbook.md#bad-makefiles should be changed, bad makefiles will probably affect both homebrew and linuxbrew so I'd assume they be reported upstream.

I'm also iffy on this one: https://github.com/Homebrew/linuxbrew/blob/5d1ef0f4a670cccb5835327bfbfede40cbb468eb/share/doc/homebrew/Formula-Cookbook.md#basic-instructions, should that point to both trackers to check if a bug has already been reported?

And I would assume the same question as above would go to which bug tracker the manpage would link to: https://github.com/Homebrew/linuxbrew/blob/f765cd358e3608cdd3088b71b16868489b73f94e/Library/Homebrew/manpages/brew.1.md#bugs

iggyvolz:
Actually, could the Formula Cookbook be tossed, and just replaced with "give new softwares to either Homebrew/Homebrew or Homebrew/Homebrew-dupes if they are pre-installed on Mac"?

sjackman:
Perhaps, but I'm disinclined to replace all of `share/doc/homebrew/*` with stubs pointing to upstream Homebrew. I'm not so concerned that all the documentation be carefully edited to replace Mac terminology with Linux terminology and replace Homebrew links with Linuxbrew links. I'm just not sure it's worth the effort. I'm inclined to do it on a case-base-case basis if a particular piece of documentation causes extreme confusion and havoc with users documented by issues opened by users.

sjackman:
Fixed.

PR #416, opened 2015-05-27
Fix cgdb formula for linuxbrew.

DoomHammer:
Readline on Linux requires to be linked with appropriate version of NCurses.

As for cgdb it requires flex which may not be available for all distros. Also, its configure script contained an error which led to searching for help2man in incorrect paths (specifically: first directory from PATH variable was malformed).

### DIFF
--- a/Library/Formula/cgdb.rb
+++ b/Library/Formula/cgdb.rb
@@ -18,8 +18,13 @@ class Cgdb < Formula
     depends_on "automake" => :build
   end
 
+  if OS.linux?
+    depends_on "flex" => :build
+  end
   depends_on "help2man" => :build
   depends_on "readline"
+
+  patch :DATA
 
   def install
     system "sh", "autogen.sh" if build.head?
@@ -29,3 +34,15 @@ class Cgdb < Formula
     system "make install"
   end
 end
+__END__
+--- a/configure	2014-11-13 19:59:03.000000000 +0000
++++ b/configure	2015-05-27 13:04:58.815408757 +0000
+@@ -5443,7 +5443,7 @@ else
+   ac_cv_prog_HAS_HELP2MAN="$HAS_HELP2MAN" # Let the user override the test.
+ else
+ as_save_IFS=$IFS; IFS=$PATH_SEPARATOR
+-for as_dir in path =$PATH
++for as_dir in $PATH
+ do
+   IFS=$as_save_IFS
+   test -z "$as_dir" && as_dir=.
--- a/Library/Formula/cgdb.rb
+++ b/Library/Formula/cgdb.rb
@@ -18,9 +18,7 @@ class Cgdb < Formula
     depends_on "automake" => :build
   end
 
-  if OS.linux?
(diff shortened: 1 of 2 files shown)


Library/Formula/cgdb.rb
```diff
@@ -18,9 +18,14 @@ class Cgdb < Formula
     depends_on "automake" => :build
   end
 
+  if OS.linux?
+    depends_on "flex" => :build
+  end
```
sjackman: Please use
```ruby
depends_on "flex" => :build unless OS.mac?
```

Library/Formula/cgdb.rb
```diff
@@ -18,9 +18,14 @@ class Cgdb < Formula
...
   depends_on "help2man" => :build
   depends_on "readline"
 
+  patch :DATA
```
sjackman: Have you submitted this patch upstream?

Library/Formula/readline.rb
```diff
@@ -33,7 +33,7 @@ class Readline < Formula
 
   def install
     ENV.universal_binary
-    system "./configure", "--prefix=#{prefix}", "--enable-multibyte"
+    system "./configure", "--prefix=#{prefix}", "--enable-multibyte", ("--with-curses" if OS.linux?)
     system "make", "install"
```
sjackman: Does this patch fix an error message that you observed? If so, can you report the error message?
Please submit this `readline` chunk in a separate PR.

Library/Formula/cgdb.rb
DoomHammer: To tell the truth I didn't know what is the proper way to fix Linux-related issues. Is it homebrew/homebrew first and then it gets merged to linuxbrew or the other way round. Shall I split this and submit to homebrew/homebrew?

Library/Formula/readline.rb
DoomHammer: The error message itself was rather cryptic and occured while configure'ing cgdb. Something along the lines of "unable to check version of readline". Actually, in config.log there was more precise message that showed missing ncurses symbols during linking stage.

Readline itself builds and installs fine even without this patch but is simply unusable.

Library/Formula/cgdb.rb
sjackman: Fixes specifically for Linux that don't affect Mac should be submitted to Linuxbrew. I should have bene more specific. I meant, have you submitted this patch upstream to the maintainers of cgdb?

DoomHammer: I haven't submitted it upstream. I tested cgdb from master and it doesn't have this issue. I believe it is only relevant to the released tarball.

sjackman: In that case, just ask the upstream authors if they'd be willing to tag a new stable release.

sjackman:
I've added the flex dependency, but not the patch, as it sounds as though it's fixed upstream.

sjackman:

PR #417, opened 2015-05-27
Binary resources for Linux in rust.

DoomHammer:
Add precompiled rust binaries for x64 Linux hosts.


sjackman:
Merged. Thanks for your contribution, Piotr.

PR #418, opened 2015-05-27
Fix readline formula for linuxbrew by linking ncurses.

DoomHammer:
Readline builds and installs fine on Linux. However, when you install a dependent application (like cgdb) you might get a cryptic error message that configure is unable to check the version of readline. What actually happens is linker failing to find symbols from ncurses that should be linked with readline.

This PR fixes the issue.

### DIFF
--- a/Library/Formula/readline.rb
+++ b/Library/Formula/readline.rb
@@ -33,6 +33,7 @@ class Readline < Formula
 
   def install
     ENV.universal_binary
+    # Add --with-curses to fix the error: CGDB requires GNU readline 5.1 or greater to link.
     system "./configure", "--prefix=#{prefix}", "--enable-multibyte", ("--with-curses" if OS.linux?)
     system "make", "install"
 
--- a/Library/Formula/readline.rb
+++ b/Library/Formula/readline.rb
@@ -33,7 +33,7 @@ class Readline < Formula
 
   def install
     ENV.universal_binary
-    system "./configure", "--prefix=#{prefix}", "--enable-multibyte"
+    system "./configure", "--prefix=#{prefix}", "--enable-multibyte", ("--with-curses" if OS.linux?)
     system "make", "install"
 
     # The 6.3 release notes say:


sjackman:
Please report the exact error message. You can use `brew gist-logs cgdb`

DoomHammer:
Here are the logs: https://gist.github.com/anonymous/4a1a3a811617708cdbea

sjackman:
There's no error message on that logfile about readline. Only flex.
https://gist.github.com/anonymous/4a1a3a811617708cdbea#file-01-configure-L136
```
configure: error: Please install flex before installing
```

DoomHammer:
Terribly sorry. It was late and I was tired. Here are the correct ones: https://gist.github.com/anonymous/8382193e9ec614a073f5

sjackman:
https://gist.github.com/anonymous/8382193e9ec614a073f5#file-01-configure-L142-L146
```
checking version of installed readline library... configure: WARNING: Could not test version of installed readline library.
configure: error: CGDB requires GNU readline 5.1 or greater to link.
  If you used --with-readline instead of using the system readline library,
  make sure to set the correct readline library on the linker search path
  via LD_LIBRARY_PATH or some other facility.
 ```

File: Library/Formula/readline.rb [hunk 1]
```diff
@@ -33,7 +33,7 @@ class Readline < Formula
 
   def install
     ENV.universal_binary
-    system "./configure", "--prefix=#{prefix}", "--enable-multibyte"
+    system "./configure", "--prefix=#{prefix}", "--enable-multibyte", ("--with-curses" if OS.linux?)
```
sjackman: Please add a comment
```ruby
# Add --with-curses to fix the error: CGDB requires GNU readline 5.1 or greater to link.
```

sjackman:
Were you able to resolve this issue without needing the patch, Piotr?

DoomHammer:
Actually, I think I have closed this by mistake.

tdsmith:
This is one of a few reasons that python can't be built with readline support, I think (it spews this message: `*** WARNING: renaming "readline" since importing it failed: /home/vagrant/.linuxbrew/lib/libreadline.so.6: undefined symbol: UP`)

sjackman:
Yeah, this is a real issue, that `libreadline.so` should be linked against `libcurses.so` but isn't by default. See http://www.linuxfromscratch.org/lfs/view/6.6/chapter06/readline.html
and http://lists.gnu.org/archive/html/bug-readline/2014-03/msg00008.html
The fix is
`make install SHLIB_LIBS=-lcurses`

sjackman:
Before:
```
$ ldd /gsc/btl/linuxbrew/lib/libreadline.so.6
	linux-vdso.so.1 (0x00007fff457fd000)
	libc.so.6 => /gsc/btl/linuxbrew/lib/libc.so.6 (0x00002ba880bee000)
	/gsc/btl/linuxbrew/Cellar/glibc/2.19/lib64/ld-linux-x86-64.so.2 (0x00002ba88078d000)
```

After:
```
$ ldd /gsc/btl/linuxbrew/lib/libreadline.so.6
	linux-vdso.so.1 (0x00007fff051fd000)
	libncursesw.so.5 => /gsc/btl/linuxbrew/lib/libncursesw.so.5 (0x00002b3ffeebb000)
	libc.so.6 => /gsc/btl/linuxbrew/lib/libc.so.6 (0x00002b3fff106000)
	/gsc/btl/linuxbrew/Cellar/glibc/2.19/lib64/ld-linux-x86-64.so.2 (0x00002b3ffea5a000)
```

sjackman:
Test out 07de187a1e0871263a36ab67fb170f39f92672fe and tell me how it goes.

sjackman:
A related issue, this hack for Lua should no longer be necessary: 8324fc6b477cee32f592bf12d82b849448811c3c

PR #433, opened 2015-06-12
Suggest installing unzip in docs.

DoomHammer:
Zipped resources are unpacked with unzip. Failure to install unzip prior to running `brew install` may cause failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,13 +35,13 @@ Paste at a Terminal prompt:
 ### Debian or Ubuntu
 
 ```sh
-sudo apt-get install build-essential curl git m4 ruby texinfo libbz2-dev libcurl4-openssl-dev libexpat-dev libncurses-dev zlib1g-dev
+sudo apt-get install build-essential curl git m4 ruby texinfo libbz2-dev libcurl4-openssl-dev libexpat-dev libncurses-dev unzip zlib1g-dev
 ```
 
 ### Fedora, CentOS or Red Hat
 
 ```sh
-sudo yum groupinstall 'Development Tools' && sudo yum install curl git irb m4 ruby texinfo bzip2-devel curl-devel expat-devel ncurses-devel zlib-devel
+sudo yum groupinstall 'Development Tools' && sudo yum install curl git irb m4 ruby texinfo bzip2-devel curl-devel expat-devel ncurses-devel unzip zlib-devel
 ```
 
 ### 32-bit x86 platforms


sjackman:
There's  logic in Homebrew to install `xz` before packages that are compressed with it. The same logic does not exist for `zip`, because it's included with Mac OS. The `xz` logic should be expanded to `unzip`.

DoomHammer:
So you suggest patching those packages that use zipped resources for dependencies?

sjackman:
Something like the following, untested, not sure if the `unless` syntax is right with `when`.
```
when '.zip' then Dependency.new('unzip', tags) unless OS.mac?
```

DoomHammer:
That would *almost* work except for the fact that homebrew/dupes/unzip is needed. Somehow during unzipping it can't find it still. Something with paths, I guess but will debug it later.

sjackman:
What's the error message that you get? Does this work?
```ruby
when '.zip' then Dependency.new('homebrew/dupes/unzip', tags) unless OS.mac?
```

DoomHammer:
unzip installs successfully with dupes approach. However I got

Already downloaded: /home/linuxbrew/.cache/Homebrew/ninja--gtest-1.7.0.zip
/home/linuxbrew/.linuxbrew/Library/Homebrew/debrew.rb:11:in `raise'
ErrorDuringExecution: Failure while executing: unzip -qq /home/linuxbrew/.cache/Homebrew/ninja--gtest-1.7.0.zip

Debug shell shows that PATH is, in fact, /usr/bin:/bin I suppose this is on purpose. In download_strategy.rb there is a line:

`with_system_path { quiet_safe_system 'unzip', {:quiet_flag => '-qq'}, cached_location }`

not much of a comment though and blame goes back to the beginning.

Do you think it should be changed for Linuxbrew?

sjackman:
@xu-cheng Would you please be able to review/test this pull request?

sjackman:
> Do you think it should be changed for Linuxbrew?

Yes, I think that on Homebrew `with_system_path` should be modified to include `HOMEBREW_PREFIX/bin`.

See https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/utils.rb#L145-L151

```ruby
def with_system_path
  old_path = ENV['PATH']
  ENV['PATH'] = '/usr/bin:/bin'
  prepend_path 'PATH', HOMEBREW_PREFIX/'bin'
  yield
ensure
  ENV['PATH'] = old_path
end
```

DoomHammer:
Just created #436 with those changes.

PR #435, opened 2015-06-13
git: tcl-tk requires full name

dunn:
Upstream changed how taps are handled so formula from, e.g., dupes, need to be given with their full name.

### DIFF
--- a/Library/Formula/git.rb
+++ b/Library/Formula/git.rb
@@ -35,7 +35,7 @@ class Git < Formula
   depends_on "curl" if build.with? "brewed-curl"
   depends_on "go" => :build if build.with? "persistent-https"
   depends_on "expat" unless OS.mac?
-  depends_on "tcl-tk" unless OS.mac?
+  depends_on "homebrew/dupes/tcl-tk" unless OS.mac?
 
   # Trigger an install of swig before subversion, as the "swig" doesn't get pulled in otherwise
   # See https://github.com/Homebrew/homebrew/issues/34554


sjackman:
This exact change was reverted in https://github.com/Homebrew/linuxbrew/commit/c7b6b170c402b2fc7b1e14ab1f8db51f994ee8a9. Read the discussion there. The problem is in bootstrapping a Linuxbrew system. The `git` formula cannot refer to a tap, like `homebrew/dupes`, because git is required to tap that tap, which results in a circular dependency. The work around is to install `tcl-tk` using http like so:
```
brew install https://raw.githubusercontent.com/Homebrew/homebrew-dupes/master/tcl-tk.rb --without-tk
brew install git
```
See https://github.com/sjackman/docker-linuxbrew/blob/master/linuxbrew-standalone/Dockerfile#L5

@xu-cheng Recommended copying the `tcl-tk` formula from `homebrew/dupes` to `homebrew/linuxbrew`. I think that's the best solution.

iggyvolz:
Couldn't we just auto-tap dupes, by downloading the Homebrew/homebrew-dupes tarball in the install file?  It would be a bit of a task to do though - I take a whack at it over the next two days but then I'm on vacation until the first week of July.

DomT4:
What's the `git` issue here? Does the `git` via `apt-get` not work nicely with `brew`?

iggyvolz:
My understanding was that the ultimate goal of linuxbrew was that you wouldn't ever need to touch apt-get.  It's a lot more convenient to install git via linuxbrew than to have to install git in order to install linuxbrew in order to install git.

dunn:
Should I open a PR for a linuxbrew version of `tcl-tk`?

DomT4:
> It's a lot more convenient to install git via linuxbrew than to have to install git in order to install linuxbrew in order to install git.

Sure, agreed. It just lists `git` as an `apt-get` requirement on the `README`, which was my confusion.

Maybe write up some logic that dictates if you want to install a formula that requires usage of a tap (i.e. from `dupes`) you must `brew install git` first? @mistydemeo did something similar in Tigerbrew with `sha256` (OpenSSL or the `sha2` utility).

iggyvolz:
@DomT4 The problem is that you can't install git until you install tcl-tk, and you can't install tcl-tk unless you tap dupes, and you can't tap unless you first install git somehow.

DomT4:
Heh, fun.

Given Linuxbrew only supports 64-bit platforms, theoretically you could vendor a `git` binary/package in `Library/Homebrew/vendor` and use that by default for all internal `git clone`s? Unless the `tcl-tk` usage is runtime?

A slimmed down vendored `git` would only be another 27MB or so.

mistydemeo:
Disabling gitk by default is a reasonable compromise, I think?

sjackman:
If `gitk` is the only reason that `git` depends on `tcl-tk` then, heck yes! Thanks for the tip, Misty.

mistydemeo:
Yeah, judging from the makefile, it appears to be solely for providing gitk:

```make
# Define NO_TCLTK if you do not want Tcl/Tk GUI.
#
# The TCL_PATH variable governs the location of the Tcl interpreter
# used to optimize git-gui for your system.  Only used if NO_TCLTK
# is not set.  Defaults to the bare 'tclsh'.
#
# The TCLTK_PATH variable governs the location of the Tcl/Tk interpreter.
# If not set it defaults to the bare 'wish'. If it is set to the empty
# string then NO_TCLTK will be forced (this is used by configure script).
```

sjackman:
`tcl-tk` is needed for `gitk` `git-citool` `git-gui `git-gui--askpass`

sjackman:
Fixed in cb035914d0383cac93ee85220120e84418766ccf "git 2.4.4: Add --without-tcl-tk for Linuxbrew"

PR #436, opened 2015-06-14
Add unzip as a dependency for zipped resources.

DoomHammer:
For systems without unzip installed by default this ensures all resources can be handled properly.

### DIFF
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -189,6 +189,7 @@ class DependencyCollector
     when '.lz'  then Dependency.new('lzip', tags)
     when '.rar' then Dependency.new('unrar', tags)
     when '.7z'  then Dependency.new('p7zip', tags)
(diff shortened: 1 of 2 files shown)


sjackman:
Thanks, Piotr. Nice work.

sjackman:
Hi, Piotr. I merge this PR, but it causes an error. Did it work for you?

```
Error: undefined method `prepend_path' for #<Object:0x0000001322bee8>
```

sjackman:
I fixed it with this patch:
```diff
-  ENV.prepend_path 'PATH', HOMEBREW_PREFIX/'bin' unless OS.mac?
+  ENV['PATH'] = "#{HOMEBREW_PREFIX}/bin:#{ENV['PATH']}" unless OS.mac?
```

DoomHammer:
It did work for me on Docker image. Strange.

sjackman:
I don't know why `ENV.prepend_path` would not be defined. Strange indeed.

PR #445, opened 2015-06-20
Update libtool.rb

digiter:
On linux we don't need to add prefix 'g'. Which fails me when brewing neovim.


sjackman:
Thanks for your contribution, @digiter.

Issue #446, opened 2015-06-21
openssl failed to install on Linux Ubuntu 15.04 (Vivid Vervet)


UnsungHero97:
I have a server on DigitalOcean running Ubuntu 15, and I'm trying to install OpenSSL, but it's failing:

    hristo@test:~$ brew install openssl
    ==> Downloading https://www.openssl.org/source/openssl-1.0.2c.tar.gz
    Already downloaded: /home/hristo/.cache/Homebrew/openssl-1.0.2c.tar.gz
    ==> perl ./Configure --prefix=/home/hristo/.linuxbrew/Cellar/openssl/1.0.2c --openssldir=/home/hristo/.linuxbrew/etc/openssl no-ssl2 zlib-dynamic shared enable-cms linux-x86_64
    ==> make depend
    ==> make
    Makefile:88: recipe for target 'subdirs' failed
    make[1]: *** [subdirs] Error 1
    make[1]: Leaving directory '/tmp/openssl20150620-4632-500bkz/openssl-1.0.2c/crypto'
    Makefile:282: recipe for target 'build_crypto' failed
    make: *** [build_crypto] Error 1

    READ THIS: https://github.com/Homebrew/linuxbrew/blob/master/share/doc/homebrew/Troubleshooting.md#troubleshooting

    These open issues may also help:
    openssl fails "make test" on 10.10.3 (https://github.com/Homebrew/homebrew/issues/40892)
    Installing Elixir fails due to OpenSSL errors. (https://github.com/Homebrew/homebrew/issues/33218)

One thing that is confusing is why the output shows "openssl fails "make test" on 10.10.3"... is the install thinking it's running on OS X and not Linux?

I've created a gist with all the output from running the install: 
https://gist.github.com/UnsungHero97/981e233b2278ba369bd1

What am I missing?

iggyvolz:
Try running brew install zlib, if it's not already installed.

UnsungHero97:
Here's what I got:

    hristo@test:~$ brew install zlib
    Error: No available formula for zlib 
    Searching formulae...
    lzlib
    Searching taps...
    homebrew/dupes/zlib

DoomHammer:
homebrew/dupes/zlib should be the one you are looking for.

iggyvolz:
brew install homebrew/dupes/zlib.  I always tap homebrew/dupes so I always
forget it's there.

UnsungHero97:
nice! that worked. thanks!

[...]

PR #447, opened 2015-06-22
Openssl depends on homebrew/dupes/zlib

iggyvolz:
See #446


sjackman:
Thanks for the fix, Iggy. I needed to removed the `homebrew/dupes` and change the patch to `depends_on "zlib" unless OS.mac?`, otherwise it creates a cyclic dependency.

sjackman:
See https://github.com/Homebrew/linuxbrew/pull/435 for the discussion.

Issue #441, opened 2015-06-17
mcrypt fails with "checking for libmcrypt - version >= 2.5.0... no"


iggyvolz:
I'm trying to brew install mcrypt, but it fails with:
```
checking for libmcrypt-config... /home/ryan/.linuxbrew/Cellar/mcrypt/2.6.8/bin/libmcrypt-config
checking for libmcrypt - version >= 2.5.0... no
*** Could not run libmcrypt test program, checking why...
*** The test program compiled, but did not run. This usually means
*** that the run-time linker is not finding LIBMCRYPT or finding the wrong
*** version of LIBMCRYPT. If it is not finding LIBMCRYPT, you'll need to set your
*** LD_LIBRARY_PATH environment variable, or edit /etc/ld.so.conf to point
*** to the installed location  Also, make sure you have run ldconfig if that
*** is required on your system
***
*** If you have an old version installed, it is best to remove it, although
*** you may also be able to get things to work by modifying LD_LIBRARY_PATH
***
configure: error: *** libmcrypt was not found
```

https://gist.github.com/anonymous/aee2769d518ede109356

I haven't the slightest idea why it's failing.  libmcrypt-config --version is returning 2.5.8 as it should, but that's not coming across to mcrypt for some reason.

sjackman:
Try
```sh
LD_LIBRARY_PATH=`brew --prefix mcrypt`/lib brew install mcrypt

**Comment by iggyvolz** [comment]
Works like a charm!

So to fix this in the linuxbrew core, would I add:
```ld_library_path = "#{prefix}/lib"```
to mcrypt within the install step?

**Comment by sjackman** [comment]
Great. Glad to hear it. Please send a pull request if this works.
```ruby
# Fix configure: error: *** libmcrypt was not found
ENV["LD_LIBRARY_PATH"] = lib
```

iggyvolz:
That worked perfectly!  I actually ran into the same exact problem setting up a server and that line made it work.  I'll send in a PR as soon as I'm done with the server.

sjackman:
Great. Thanks, Iggy.

PR #448, opened 2015-06-22
Mcrypt must have LD_LIBRARY_PATH set on install

iggyvolz:
See #441 

I'm pretty sure I did this right, but my ruby's not advanced by any means.


sjackman:
Works for me. Thanks, Iggy.

Issue #453, opened 2015-06-26
The default-names options is not respected in libtool formula


stanley-cheung:
May be related to #445 

```
stanleycheung@stanleycheung-test:~/homebrew-grpc$ brew install libtool
==> Downloading http://ftpmirror.gnu.org/libtool/libtool-2.4.6.tar.xz
Already downloaded: /home/stanleycheung/.cache/Homebrew/libtool-2.4.6.tar.xz
==> ./configure --prefix=/home/stanleycheung/.linuxbrew/Cellar/libtool/2.4.6  --enable-ltdl-install
==> make install
==> Caveats
In order to prevent conflicts with Apple's own libtool we have prepended a "g"
so, you have instead: glibtool and glibtoolize.


==> Summary
/home/stanleycheung/.linuxbrew/Cellar/libtool/2.4.6: 69 files, 3.9M, built in 5 seconds
```

Notice that the `--program-prefix=g` flag is not being added to the `./configure` command.

Same behavior if I run `brew install libtool --without-default-names`

Interestingly, the option works if I rename it to `bla` or `default-name`. It only stops being respected when it's named `default-names`.

(Edit: To be more specific, if I rename the option to `with-bla`, and run `brew install libtool` or `brew install libtool --without-bla`, the `--program-prefix=g` option will be added to the `./configure` command. If I run `brew install libtool --with-bla`, the `--program-prefix=g` option will NOT be added. So all these are expeced behavior. However, if the option is named `with-default-names`, the `--program-prefix=g` option is NOT added regardless of how I invoke `brew install libtool`. Hope it's clear.)

It's breaking our formula because the `autoconf` formula is still expecting the binary to be called `glibtoolize`.

sjackman:
There's magic in Linuxbrew that adds `--default-names` to every build. See https://github.com/Homebrew/linuxbrew/blob/37f6ab9c5da46c9e97a1e298a0094f537ca71362/Library/Homebrew/build_options.rb#L8-L10

I'd accept a pull request to add support for `--without-default-names`, which is currently missing.

sjackman:
The program is usually called `libtoolize` on Linux not `glibtoolize`. Which formula is breaking? I'd accept a pull request to add a symlink from `libtoolize` to `glibtoolize` though.

stanley-cheung:
For example, the `autoconf` formula is looking for the `glibtoolize` binary.

[...]

PR #454, opened 2015-06-26
libtool: Add symlink to libtool's old name

stanley-cheung:
* Add symlink to `libtool`'s old name `glibtool`, because some other formula, like `autoconf`, is still expecting the `glibtool` binary.
 * For issue #453


stanley-cheung:
@tbetbetbe FYI

sjackman:
Thanks for your contribution, Stanley.

PR #457, opened 2015-06-30
Fix bib-tool formula to work with Linuxbrew

### LINKED ISSUE #456: New formula for bib-tool doesn't work on linux
In the newest version of BibTool (2.60), there is a problem with the tarball. It has duplicate files. It works perfectly fine to unarchive the tarball with gnutar but not with bsdtar. The latter doesn't like duplicate files. In order to work around this, the new bib-tool formula does:

```ruby
class BibToolDownloadStrategy < CurlDownloadStrategy
  def stage
     with_system_path { safe_system 'tar', 'xqf', cached_location, 'BibTool/doc/bibtool.tex' }
     with_system_path { safe_system 'tar', 'xf', cached_location, '--exclude', 'BibTool/doc/bibtool.tex' }
     chdir
  end
end
```

That is, it ends up running:

```sh
tar xqf bib-tool-2.60.tar.gz BibTool/doc/bibtool.tex
tar xf --exclude bib-tool-2.60.tar.gz BibTool/doc/bibtool.tex
```
(See https://github.com/Homebrew/homebrew/issues/40559)

However, this is a problem with gnutar (which is what is installed on most (all?) linux machines) since `-q` (AKA `--fast-read`) is only implemented in bsdtar.

I don't know how the formula maintenance works between Homebrew and Linuxbrew, but I assume you like to keep the formulas the same, correct?

If so, something like the following will need to be done.

```ruby
class BibToolDownloadStrategy < CurlDownloadStrategy
  if `tar --version`.match 'bsdtar'
    def stage
      with_system_path { safe_system 'tar', 'xqf', cached_location, 'BibTool/doc/bibtool.tex' }
      with_system_path { safe_system 'tar', 'xf', cached_location, '--exclude', 'BibTool/doc/bibtool.tex' }
      chdir
    end
  else
    def stage
      with_system_path { safe_system 'tar', 'xf', cached_location }
      chdir
    end
  end
end
```

adamliter:
The tarball for v. 2.60 of BibTool has a problem where it contains a
duplicate file. This does not play nicely with bsdtar, which is what is
installed on Macs. See https://github.com/Homebrew/homebrew/issues/40559.

The commit that fixed that issue, however, made use of --fast-read (-q),
which is only implemented in bsdtar, not gnutar. In order to make sure
that this formula works with Linuxbrew, we need to define "stage" based
on the flavor of tar that is installed on the machine.

Closes #456.

### DIFF
--- a/Library/Formula/bib-tool.rb
+++ b/Library/Formula/bib-tool.rb
@@ -3,9 +3,13 @@
 # See https://github.com/Homebrew/homebrew/issues/40559
 class BibToolDownloadStrategy < CurlDownloadStrategy
   def stage
-     with_system_path { safe_system 'tar', 'xqf', cached_location, 'BibTool/doc/bibtool.tex' }
-     with_system_path { safe_system 'tar', 'xf', cached_location, '--exclude', 'BibTool/doc/bibtool.tex' }
-     chdir
+    if `tar --version`.match 'bsdtar'
+      with_system_path { safe_system 'tar', 'xqf', cached_location, 'BibTool/doc/bibtool.tex' }
+      with_system_path { safe_system 'tar', 'xf', cached_location, '--exclude', 'BibTool/doc/bibtool.tex' }
+    else
+      with_system_path { safe_system 'tar', 'xf', cached_location }
+    end
+    chdir
   end
 end
 


sjackman:
Thanks for your contribution, Adam.

adamliter:
No problem! Sorry that I didn't think of this initially when I opened the pull request with the weird workaround for the problematic tarball.

sjackman:
No worries. An alternative workaround, if this comes up again, which I don't expect, would be to simply repackage the tarball and host it on Bintray.

adamliter:
Yes, hopefully it won't be a problem after the next release.

PR #460, opened 2015-07-02
python: hint dependencies

### LINKED ISSUE #420: python - ImportError: No module named '_ssl'
Doing a fresh install of everything, getting critical errors trying to use python:
(openssl is installed - brew and system)

```
% brew install python3
% pip
Traceback (most recent call last):
  File "/bio/linuxbrew/bin/pip", line 9, in <module>
    load_entry_point('pip==6.1.1', 'console_scripts', 'pip')()
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 552, in load_entry_point
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 2672, in load_entry_point
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 2345, in load
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 2351, in resolve
  File "/bio/linuxbrew/Cellar/python/2.7.10/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/__init__.py", line 15, in <module>
    from pip.vcs import git, mercurial, subversion, bazaar  # noqa
  File "/bio/linuxbrew/Cellar/python/2.7.10/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/vcs/mercurial.py", line 10, in <module>
    from pip.download import path_to_url
  File "/bio/linuxbrew/Cellar/python/2.7.10/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/download.py", line 32, in <module>
    from pip._vendor import requests, six
  File "/bio/linuxbrew/Cellar/python/2.7.10/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/_vendor/__init__.py", line 92, in load_module
    raise ImportError("No module named '%s'" % (name,))
ImportError: No module named 'pip._vendor.requests'
```

```
% brew install python3
% pip3

Traceback (most recent call last):
  File "/bio/linuxbrew/bin/pip3", line 9, in <module>
    load_entry_point('pip==6.1.1', 'console_scripts', 'pip3')()
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/setuptools-15.2-py3.4.egg/pkg_resources/__init__.py", line 552, in load_entry_point
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/setuptools-15.2-py3.4.egg/pkg_resources/__init__.py", line 2672, in load_entry_point
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/setuptools-15.2-py3.4.egg/pkg_resources/__init__.py", line 2345, in load
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/setuptools-15.2-py3.4.egg/pkg_resources/__init__.py", line 2351, in resolve
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/__init__.py", line 15, in <module>
    from pip.vcs import git, mercurial, subversion, bazaar  # noqa
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/vcs/mercurial.py", line 10, in <module>
    from pip.download import path_to_url
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/download.py", line 32, in <module>
    from pip._vendor import requests, six
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/_vendor/requests/__init__.py", line 59, in <module>
    from .models import Request, Response, PreparedRequest
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/_vendor/requests/models.py", line 19, in <module>
    from .packages.urllib3.fields import RequestField
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/_vendor/requests/packages/urllib3/__init__.py", line 10, in <module>
    from .connectionpool import (
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/_vendor/requests/packages/urllib3/connectionpool.py", line 31, in <module>
    from .connection import (
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/_vendor/requests/packages/urllib3/connection.py", line 45, in <module>
    from .util.ssl_ import (
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/_vendor/requests/packages/urllib3/util/__init__.py", line 5, in <module>
    from .ssl_ import (
  File "/bio/linuxbrew/opt/python3/lib/python3.4/site-packages/pip-6.1.1-py3.4.egg/pip/_vendor/requests/packages/urllib3/util/ssl_.py", line 12, in <module>
    import ssl
  File "/bio/linuxbrew/opt/python3/lib/python3.4/ssl.py", line 97, in <module>
    import _ssl             # if we can't import it, let the error propagate
ImportError: No module named '_ssl'
```

tdsmith:
Add include paths to CPPFLAGS and library paths to LDFLAGS so extension modules find the Homebrew libraries on which they depend.

Fixes #420.

- [x] Test on Trusty
- [x] Test on CentOS 7
- [x] Apply symmetric change to python3.rb
- [x] Check whether `-I#{HOMEBREW_PREFIX}/include -L#{HOMEBREW_PREFIX}/lib` is sufficient

### DIFF
--- a/Library/Formula/python.rb
+++ b/Library/Formula/python.rb
@@ -94,7 +94,7 @@ class Python < Formula
     args << (OS.mac? ? "--enable-framework=#{frameworks}" : "--enable-shared")
     args << "--without-gcc" if ENV.compiler == :clang
 
-    unless MacOS::CLT.installed?
+    if OS.mac? && !MacOS::CLT.installed?
       # Help Python's build system (setuptools/pip) to build things on Xcode-only systems
       # The setup.py looks at "-isysroot" to get the sysroot (and not at --sysroot)
       cflags = "CFLAGS=-isysroot #{MacOS.sdk_path}"
@@ -106,6 +106,10 @@ class Python < Formula
       end
       args << cflags
       args << ldflags
+    elsif OS.linux?
+      cppflags = "CPPFLAGS=-I#{HOMEBREW_PREFIX}/include"
+      ldflags = "LDFLAGS=-L#{HOMEBREW_PREFIX}/lib"
+      args << cppflags << ldflags
     end
 
     # Avoid linking to libgcc https://code.activestate.com/lists/python-dev/112195/
--- a/Library/Formula/python.rb
+++ b/Library/Formula/python.rb
@@ -119,7 +119,7 @@ class Python < Formula
     # superenv handles that cc finds includes/libs!
     inreplace "setup.py" do |s|
       s.gsub! "do_readline = self.compiler.find_library_file(lib_dirs, 'readline')",
-              "do_readline = '#{Formula["readline"].opt_lib}/libhistory.dylib'"
+              "do_readline = '#{Formula["readline"].opt_lib}/libhistory.#{OS.mac? ? "dylib" : "so"}'"
       s.gsub! "/usr/local/ssl", Formula["openssl"].opt_prefix
     end
 
--- a/Library/Formula/python.rb
+++ b/Library/Formula/python.rb
@@ -28,6 +28,7 @@ class Python < Formula
   depends_on "gdbm" => :recommended
   depends_on "openssl"
   depends_on "homebrew/dupes/tcl-tk" => :optional
+  depends_on "homebrew/dupes/ncurses" => :recommended if OS.linux?
   depends_on :x11 if build.with?("tcl-tk") && Tab.for_name("homebrew/dupes/tcl-tk").with?("x11")
 
   skip_clean "bin/pip", "bin/pip-2.7"
@@ -108,6 +109,7 @@ class Python < Formula
       args << ldflags
     elsif OS.linux?
       cppflags = "CPPFLAGS=-I#{HOMEBREW_PREFIX}/include"
+      cppflags << " -I#{Formula["ncurses"].opt_include}/ncursesw" if build.with?("ncurses")
       ldflags = "LDFLAGS=-L#{HOMEBREW_PREFIX}/lib"
       args << cppflags << ldflags
     end
--- a/Library/Formula/python3.rb
+++ b/Library/Formula/python3.rb
@@ -100,7 +100,7 @@ class Python3 < Formula
     args << (OS.mac? ? "--enable-framework=#{frameworks}" : "--enable-shared")
     args << '--without-gcc' if ENV.compiler == :clang
 
-    unless MacOS::CLT.installed?
+    if OS.mac? && !MacOS::CLT.installed?
       # Help Python's build system (setuptools/pip) to build things on Xcode-only systems
       # The setup.py looks at "-isysroot" to get the sysroot (and not at --sysroot)
       cflags = "CFLAGS=-isysroot #{MacOS.sdk_path}"
@@ -111,7 +111,12 @@ class Python3 < Formula
       end
       args << cflags
       args << ldflags
+    elsif OS.linux?
+      cppflags = "CPPFLAGS=-I#{HOMEBREW_PREFIX}/include"
+      ldflags = "LDFLAGS=-L#{HOMEBREW_PREFIX}/lib"
+      args << cppflags << ldflags
     end
+
     # Avoid linking to libgcc http://code.activestate.com/lists/python-dev/112195/
     args << "MACOSX_DEPLOYMENT_TARGET=#{MacOS.version}"
 
--- a/Library/Formula/python3.rb
+++ b/Library/Formula/python3.rb
@@ -32,6 +32,7 @@ class Python3 < Formula
   depends_on "openssl"
   depends_on "xz" => :recommended  # for the lzma module added in 3.3
   depends_on "homebrew/dupes/tcl-tk" => :optional
+  depends_on "homebrew/dupes/ncurses" => :recommended if OS.linux?
   depends_on :x11 if build.with? "tcl-tk" and Tab.for_name("homebrew/dupes/tcl-tk").with? "x11"
 
   skip_clean "bin/pip3", "bin/pip-3.4", "bin/pip-3.5"
@@ -113,6 +114,7 @@ class Python3 < Formula
       args << ldflags
     elsif OS.linux?
       cppflags = "CPPFLAGS=-I#{HOMEBREW_PREFIX}/include"
+      cppflags << " -I#{Formula["ncurses"].opt_include}/ncursesw" if build.with?("ncurses")
       ldflags = "LDFLAGS=-L#{HOMEBREW_PREFIX}/lib"
       args << cppflags << ldflags
     end
--- a/Library/Formula/python3.rb
+++ b/Library/Formula/python3.rb
@@ -124,7 +124,7 @@ class Python3 < Formula
     # superenv makes cc always find includes/libs!
     inreplace "setup.py" do |s|
       s.gsub! "do_readline = self.compiler.find_library_file(lib_dirs, 'readline')",
-              "do_readline = '#{Formula["readline"].opt_lib}/libhistory.dylib'"
+              "do_readline = '#{Formula["readline"].opt_lib}/libhistory.#{OS.mac? ? "dylib" : "so"}'"
       s.gsub! "/usr/local/ssl", Formula["openssl"].opt_prefix
     end
 


tdsmith:
- [x] Test on Trusty
- [ ] Test on CentOS 7
- [ ] Apply symmetric change to python3.rb

tdsmith:
Ping @sjackman for review; this works, but I don't know much about your preferred style for Linuxbrew changes. Let me know if anything needs tweaking.

tseemann:
@tdsmith @sjackman  

 I can confirm that your modified recipe works on RHEL 7.x for python 2 and 3 (Hooray!)

sjackman:
Thanks for this, Tim. Does compiling Python work out-of-the-box using a Homebrewed `gcc`? Does this patch support primarily using the system's `/usr/bin/gcc`?

Library/Formula/python.rb
```diff
@@ -106,6 +106,17 @@ def install
...
+      hints = ["openssl"]
+      hints << "readline" if build.with? "readline"
+      hints << "sqlite" if build.with? "sqlite"
+      hints << "gdbm" if build.with? "gdbm"
+      hints << "tcl-tk" if build.with? "tcl-tk"
+      hint_includes = hints.map { |f| "-I#{Formula[f].opt_include}" }
+      hint_libs = hints.map { |f| "-L#{Formula[f].opt_lib}" }
+      cppflags = "CPPFLAGS=#{hint_includes.join " "}"
+      ldflags = "LDFLAGS=#{hint_libs.join " "}"
+      args << cppflags << ldflags
```
sjackman: Does `-I#{HOMEBREW_PREFIX}/include -L#{HOMEBREW_PREFIX}/lib` work?

Library/Formula/python.rb
```diff
@@ -115,7 +126,7 @@ def install
     # superenv handles that cc finds includes/libs!
     inreplace "setup.py" do |s|
       s.gsub! "do_readline = self.compiler.find_library_file(lib_dirs, 'readline')",
-              "do_readline = '#{Formula["readline"].opt_lib}/libhistory.dylib'"
+              "do_readline = '#{Formula["readline"].opt_lib}/libhistory.so'"
```
sjackman: Do you get an error message without this fix? What is the effect of leaving this as is?

sjackman:
> I can confirm that your modified recipe works on RHEL 7.x for python 2 and 3

Good to know! I'm very curious as to why though, because Homebrew already adds to the command line
`-I#{HOMEBREW_PREFIX}/include -L#{HOMEBREW_PREFIX}/lib`

sjackman:
Oh, I think I see now. It looks like it's because the `python` formula is completely overriding the standard Homebrew definition of `CPPFLAGS`, `CFLAGS` and `LDFLAGS`. Is that correct?

tdsmith:
I just saw your comment about errors, Torsten; I'm guilty of this too, but can you add a new comment instead of editing when the content substantially changes? A new comment triggers an email notification but an edit doesn't.

> I can't seem to pip install things that need to compile eg. numpy (needs -dev stuff) and also it seems to be using system BLAS. Is this why I'm mean to use homebrew-python/numpy ?

Yes. Our Python configuration exposes a bug in numpy's custom distutils that we need to patch around that yields that error message about headers. The fix has been accepted upstream but not released. :( It's only relevant when building from source, which most OS X users don't do because pypi hosts wheels. As far as which BLAS is used, that's something you have to configure at numpy's build time, which the Homebrew formula takes care of.

> Does compiling Python work out-of-the-box using a Homebrewed gcc?

I haven't tested.

> Does this patch support primarily using the system's /usr/bin/gcc?

Yes.

> Oh, I think I see now. It looks like it's because the python formula is completely overriding the standard Homebrew definition of CPPFLAGS, CFLAGS and LDFLAGS. Is that correct?

It may actually be doing that now but it was not doing that before; Python's hacky setup.py, which builds the compiled extensions, parses the FLAGS variables to learn search paths for the dependencies of the compiled extension modules.

> Homebrew already adds to the command line `-I#{HOMEBREW_PREFIX}/include`

If you mean literally that Homebrew is wrapping compiler invocations and appending this to the argument list, that won't help, because setup.py places the discovered include path earlier on the command line.

Library/Formula/python.rb
tdsmith: No, but Python fails to detect whether readline is linked against curses or not, which can cause the _curses module to misbuild.

Library/Formula/python.rb
tdsmith: It may; I hadn't appreciated that these formulas aren't keg-only on Linux. I'll try that later.

Library/Formula/python.rb
tdsmith: Fixed the style.

sjackman:
>> Homebrew already adds to the command line -I#{HOMEBREW_PREFIX}/include

> If you mean literally that Homebrew is wrapping compiler invocations and appending this to the argument list, that won't help, because setup.py places the discovered include path earlier on the command line.

Homebrew adds the `-I#{HOMEBREW_PREFIX}/include` option to `CPPFLAGS` and `-L#{HOMEBREW_PREFIX}/lib` to `LDFLAGS`.

Linuxbrew's `gcc` adds `-L#{HOMEBREW_PREFIX}/lib` to the [`specs` file](https://github.com/Homebrew/linuxbrew/blob/master/Library/Formula/gcc.rb#L266-L270), which precedes the values specified on the command line. I only just noticed that it does *not* add `-I#{HOMEBREW_PREFIX}/include` to the `specs` file. It really should! That may be the ultimate source of a lot of our Python woes.

sjackman:
> because setup.py places the discovered include path earlier on the command line.

That's terrible.

tdsmith:
Ah: Homebrew uses `-isystem`, not `-I`.
```
[vagrant@localhost Python-2.7.10]$ echo $CPPFLAGS
-isystem/home/vagrant/.linuxbrew/include
```
setup.py ignores `-i`. Adding `-I` is sufficient.

Updated to just hint the Linuxbrew prefix's include and library paths. Also added a fix for building the _curses module, which I noticed was failing.

FWIW, I've received some strenuous feedback that prioritizing system search paths ahead of user-specified search paths is undesirable, e.g. here: https://github.com/Homebrew/homebrew/issues/26272#issuecomment-96406197 -- it's hard to accommodate that in Python in stdenv without fixing up the memorized configuration variables.

sjackman:
@zpincus wrote https://github.com/Homebrew/homebrew/issues/26272#issuecomment-96497263
> if that 3rd-party module contains a C extension that has a private header with the same name as any header file that is in /usr/local/include (or, more generally, brew --prefix). The same is true for linking libraries.

I've had this same issue outside of Python. See https://github.com/Homebrew/homebrew-science/issues/2458#issuecomment-117318345

sjackman:
The challenge is that we want to prioritize and enforce the following:
1. the package's private header/library files
2. Homebrew's header/library files
3. the system's header library files
4. and never the user's non-system header library/files

It's difficult to get 1 before 2 while still getting 2 before 3.

prologic:
This worked for me too. I also commented on #420 but I believe this PR fixes the issue at hand.

prologic:
NB: I hand patched my ``python.rb`` formula however since this PR containers merge conflicts apparently

sjackman:
Hi, Tim. Sorry for the long delay in getting around to reviewing your patch. I've applied a reduced version of this patch in cd680d7 that addresses the core issue, which is adding `-I#{HOMEBREW_PREFIX}/include` to `CPPFLAGS`. Thanks for helping troubleshoot this issue.

I haven't seen any issues filed regarding `ncurses` for Python, so I've left that chunk out for now.

PR #463, opened 2015-07-02
py2cairo 1.10.0: Fix for linuxbrew

paulhybryant:
Fix the error: No module _cairo can be found
The error is only triggered when the python module is used.

### DIFF
--- a/Library/Formula/py2cairo.rb
+++ b/Library/Formula/py2cairo.rb
@@ -29,9 +29,9 @@ class Py2cairo < Formula
 
     # disable waf's python extension mode because it explicitly links libpython
     # https://code.google.com/p/waf/issues/detail?id=1531
-    inreplace "src/wscript", "pyext", ""
-    ENV["LINKFLAGS"] = "-undefined dynamic_lookup"
-    ENV.append_to_cflags `python-config --includes`
+    inreplace "src/wscript", "pyext", "" if OS.mac?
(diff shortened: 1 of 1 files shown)


tdsmith:
Using the pyext mode is probably correct on Linux and should let you skip the renaming dance, if you'd rather go that route.

paulhybryant:
Can you elaborate more about pyext

Library/Formula/py2cairo.rb
```diff
@@ -30,7 +30,7 @@ def install
     # disable waf's python extension mode because it explicitly links libpython
     # https://code.google.com/p/waf/issues/detail?id=1531
     inreplace "src/wscript", "pyext", ""
-    ENV["LINKFLAGS"] = "-undefined dynamic_lookup"
+    ENV["LINKFLAGS"] = "-undefined dynamic_lookup" if OS.mac?
     ENV.append_to_cflags `python-config --includes`
```
tdsmith: You can probably remove these there lines (or make them conditional on `OS.mac?`) and then get rid of the `mv` below.

paulhybryant:
I have updated the change according to tdsmith's comment.
It works fine in my test.
Thanks,

sjackman:
Instead of deleting this code, please move it inside `if OS.mac?`.
Please see https://github.com/Homebrew/linuxbrew/blob/master/CONTRIBUTING.md

paulhybryant:
Sure. Updated.

sjackman:

PR #464, opened 2015-07-02
pygtk 2.24.0: Fix for linuxbrew

paulhybryant:
Fix the error: C compiler cannot create executables


sjackman:
Thanks for your contribution, Paul.

PR #465, opened 2015-07-02
vte 0.29.2: Fix for linuxbrew

paulhybryant:
Fix the error: No package x11 found

### DIFF
--- a/Library/Formula/vte.rb
+++ b/Library/Formula/vte.rb
@@ -25,6 +25,9 @@ class Vte < Formula
     # pygobject-codegen-2.0, but the vte Makefile does not detect this.
     ENV["PYGTK_CODEGEN"] = Formula["pygobject"].bin/"pygobject-codegen-2.0"
 
+    ENV.append_path "PKG_CONFIG_LIBDIR", "/usr/lib/pkgconfig"
+    ENV.append_path "PKG_CONFIG_LIBDIR", "/usr/lib/x86_64-linux-gnu/pkgconfig"
+    ENV.append_path "PKG_CONFIG_LIBDIR", "/usr/share/pkgconfig"
     system "./configure", *args
     system "make", "install"
   end
@@ -49,6 +52,7 @@ class Vte < Formula
     libpng = Formula["libpng"]
     pango = Formula["pango"]
     pixman = Formula["pixman"]
+    backend = OS.mac? ? "quartz" : "x11"
     flags = (ENV.cflags || "").split + (ENV.cppflags || "").split + (ENV.ldflags || "").split
     flags += %W[
       -I#{atk.opt_include}/atk-1.0
@@ -77,12 +81,12 @@ class Vte < Formula
       -L#{pango.opt_lib}
       -latk-1.0
       -lcairo
-      -lgdk-quartz-2.0
+      -lgdk-#{backend}-2.0
       -lgdk_pixbuf-2.0
       -lgio-2.0
       -lglib-2.0
       -lgobject-2.0
-      -lgtk-quartz-2.0
+      -lgtk-#{backend}-2.0
       -lintl
       -lpango-1.0
       -lpangocairo-1.0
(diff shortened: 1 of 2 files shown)


paulhybryant:
Please advise a better way to change the environment variable to build with system x11. Right now this is hard-coded.

sjackman:
Please report `brew environment |grep PKG`.
See https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/extend/ENV/std.rb#L27-L29
and https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/extend/ENV/std.rb#L96-L103
and https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/extend/ENV/std.rb#L250-L254

sjackman:
Also see http://ilgthegeek.livius.net/2015/02/05/pkg-config-search-path/
for an explanation of `PKG_CONFIG_PATH` and `PKG_CONFIG_LIBDIR`.

paulhybryant:
brew environment | grep PKG
returns
export PKG_CONFIG_LIBDIR="$HOME/.linuxbrew/lib/pkgconfig:/usr/lib/pkgconfig"

From https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/extend/ENV/std.rb#L96-L103
It seems /usr/lib/x86_64-linux-gnu/pkgconfig:/usr/share/pkgconfig should also be added in the env variable?

sjackman:
I'd be more include to remove system pkg-config paths from the default `PKG_CONFIG_LIBDIR`, including `/usr/lib/pkgconfig`.
https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/extend/ENV/std.rb#L101

paulhybryant:
I have updated the pull request.

sjackman:
I've merged the backend chunk. I haven't merged the `pkgconfig` chunk. I need to think more about how the `pkgconfig` search path should be handled. Thanks for your contribution, Paul.

paulhybryant:
Sure. But without the pkgconfig trunk, the build error will still be there when installing vte.

A potential solution could be to modify the code here
https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/extend/ENV/std.rb#L250-L254

The system pkgconfig path can be added here for x11

sjackman:
Possible. You can also symlink those `pkgconfig` files into `$HOMEBREW_PREFIX/lib/pkgconfig`.

sjackman:
You can also fake a `xorg` package like so:
```sh
mkdir -p `brew --prefix`/Cellar/xorg/7.7/lib/pkgconfig
ln -s /usr/lib64/pkgconfig/x*.pc `brew --prefix`/Cellar/xorg/7.7/lib/pkgconfig/
brew link xorg
```

paulhybryant:
Sure. There can be many ways to have it build successfully. It would be ideal that a formula just works out of the box.

Faking a package is new to me. Thanks for the tip :P.

sjackman:
I removed `/usr/lib/pkgconfig` from `PKG_CONFIG_LIBDIR` in 8c34ef1cab453d4d7a86119c10314b378a08190e

There's a `brew diy` for the purpose of faking packages. I always just do it the manual way. I think it would be helpful to have a command to help users' fake packages for software provided by their system, such as compilers, X, et c.

paulhybryant:
Just want to revisit the issue brought up here.

Faking a package by creating symlinks and link the fake package works, but it is inconvenient sometimes. For example, when I try to link the pkgconfigs for x11, they have their own dependencies (e.g. xcb requires pthread-stubs'). We will have to create symlinks for all its dependencies.
Appending the system pkgconfig paths to the pkgconfig ENV vars seem to be more convenient.

The brew diy command doesn't seem to be documented. Can you elaborate a bit how it works with an example?

Thanks,

sjackman:
I'm on vacation until the new year. brew diy is intended to install a tarball (that uses autoconf or cmake) that doesn't have a formula. Linxubrew is easier to debug when it doesn't use system libraries, which vary a lot between distributions. I'd like to package X11 client libraries for Linuxbrew,

PR #468, opened 2015-07-08
README: typo fix

viktorahlstrom:


README.md
```diff
@@ -101,8 +101,8 @@ Please report security issues to security@brew.sh.
 
 This is our PGP key which is valid until June 17, 2016.
 * Key ID: `CE59E297`
```
Mikaela: Why use short Key ID instead of 0xlong? I would suggest this being changed to `0xE33A3D3CCE59E297` (0x is optional).

*I know this part wasn't changed in this PR, but I only just now noticed it and it would probably be better to update it in same PR than opening new one just for it.*

viktorahlstrom:
@Mikaela Added your request to the PR.

xu-cheng:
This PR should be submit to [Homebrew](https://github.com/Homebrew/homebrew).  See [CONTRIBUTING.md](https://github.com/Homebrew/linuxbrew/blob/master/CONTRIBUTING.md) for more details on where PR should send to?

viktorahlstrom:
@xu-cheng This typo was already corrected in homebrew/Homebrew.

iggyvolz:
If that's the case, it will be merged into Linuxbrew within a week.

PR #476, opened 2015-07-10
python: Add an build option to allow building python with wide unicode support.

paulhybryant:
Add an build option to allow building python with wide unicode support.

### DIFF
--- a/Library/Formula/python.rb
+++ b/Library/Formula/python.rb
@@ -13,8 +13,10 @@ class Python < Formula
     sha256 "35c5e98031dbddf10cc02686e97c9cc8ca53429e1c299185303b49a8408516be" => :mountain_lion
   end
 
-  # Please don't add a wide/ucs4 option as it won't be accepted.
+  # Homebrew doesn't accept a wide/ucs4 option because narrow build is the de facto standard
+  # on Windows and OSX, but wide seems to be the default for linux
   # More details in: https://github.com/Homebrew/homebrew/pull/32368
+  option "with-unicode-ucs4", "Build unicode support with UCS4"'
   option :universal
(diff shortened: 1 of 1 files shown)


sjackman:
@tdsmith @xu-cheng @mistydemeo Does this patch make sense for Linuxbrew?

Library/Formula/python.rb
```diff
@@ -13,12 +13,14 @@ class Python < Formula
...
-  # Please don't add a wide/ucs4 option as it won't be accepted.
-  # More details in: https://github.com/Homebrew/homebrew/pull/32368
   option :universal
   option "with-quicktest", "Run `make quicktest` after the build (for devs; may fail)"
   option "with-tcl-tk", "Use Homebrew's Tk instead of OS X Tk (has optional Cocoa and threads support)"
   option "with-poll", "Enable select.poll, which is not fully implemented on OS X (https://bugs.python.org/issue5154)"
+  # Homebrew doesn't accept a wide/ucs4 option because narrow build is the de facto standard
+  # on Windows and OSX, but wide seems to be the default for linux
+  # More details in: https://github.com/Homebrew/homebrew/pull/32368
+  option "with-unicode-ucs4", "Build unicode support with UCS4"'
```
sjackman: Please move this chunk up so that the addition lines are directly next to the deletion lines. It helps merging and resolving merge conflicts.

tdsmith:
I think this is reasonable on Linux, especially in the absence of bottles. (I would suggest that `ucs4` become the default on Linux except that it would break everyone's installed packages, which is probably bad.)

paulhybryant:
Updated the commits according to comment.

sjackman:

PR #477, opened 2015-07-10
zsh: Depends on texinfo

paulhybryant:
Add zsh dependency texinfo.

### DIFF
--- a/Library/Formula/zsh.rb
+++ b/Library/Formula/zsh.rb
@@ -15,6 +15,7 @@ class Zsh < Formula
 
   depends_on 'gdbm'
   depends_on 'pcre'
+  depends_on 'texinfo' unless OS.mac?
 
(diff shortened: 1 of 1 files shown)


sjackman:
Please add `unless OS.mac?`
```ruby
depends_on "texinfo" unless OS.mac?`
```

Please report the exact error message that you see without using this patch.

sjackman:
The commit log message should be
`zsh: Depends on texinfo for Linuxbrew`

paulhybryant:
The error without this patch is

/bin/sh: line 3: makeinfo: command not found
make[1]: *** [install.info] Error 1
make[1]: Leaving directory `/tmp/zsh20150713-13485-19a2t7g/zsh-5.0.8/Doc'
make: *** [install.info] Error 2

paulhybryant:
The change and message are updated.

sjackman:

PR #480, opened 2015-07-13
Remove extra quote.

paulhybryant:
Accidentally put an extra quote there. Sorry :(

Tested and it works now.


sjackman:
Woops. I should have caught that.

PR #486, opened 2015-07-16
bison and flex dependancies in gobject-introspection

andrewjpage:
from https://github.com/Homebrew/homebrew/pull/41775

### DIFF
--- a/Library/Formula/gobject-introspection.rb
+++ b/Library/Formula/gobject-introspection.rb
@@ -19,6 +19,11 @@ class GobjectIntrospection < Formula
   depends_on "libffi"
   depends_on "cairo" => :build if build.with? "tests"
 
+  if OS.linux?
+    depends_on "bison"
+    depends_on "flex"
+  end
+
   # Allow tests to execute on OS X (.so => .dylib)
   patch do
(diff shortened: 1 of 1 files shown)


Library/Formula/gobject-introspection.rb
```diff
@@ -19,6 +19,11 @@ class GobjectIntrospection < Formula
   depends_on "libffi"
   depends_on "cairo" => :build if build.with? "tests"
 
+  if OS.linux?
+    depends_on "bison"
+    depends_on "flex"
+  end
+
```
sjackman: Please change this to
```ruby
depends_on "bison" unless OS.mac?
depends_on "flex" unless OS.mac?
```
and delete the blank line.

sjackman:
Please report the exact error message that you see without this fix, so that the issue can be searched by error message.

andrewjpage:
The errors are:
```
checking for flex... no
checking for lex... no
configure: error: flex not found but required
```
and 
```
checking for bison... no
configure: error: bison not found but required
```

sjackman:
Merged. Thanks for your contribution, Andrew.

Please squash multiple commits into a single commit. I changed the commit log message to
```
gobject-introspection: Fix depends for Linuxbrew

Depends on bison and flex.

Fix
configure: error: flex not found but required
configure: error: bison not found but required
```

sjackman:
Please keep the first line to fewer than 50 characters. Please include the error message in the commit log message.

PR #490, opened 2015-07-21
kubernetes-cli: linux support

sublimino:


Library/Formula/kubernetes-cli.rb
```diff
@@ -15,11 +15,12 @@ class KubernetesCli < Formula
   depends_on "go" => :build
 
   def install
+    os = if OS.linux? then 'linux' else 'darwin' end
```
sjackman: Please use `?` and `:` and double quotes, like the line below. See the output of `brew audit --strict kubernetes-cli`

sublimino:
I get exit code 0 from `brew audit --strict kubernetes-cli`, but the `brew tests` suite also fails for me before and after this change so it's probably something local. Could you paste the output if I need to see it please?

sjackman:
`brew audit --strict kubernetes-cli` should have suggested using `?:` rather than `if then else end`. It looks good otherwise.

sjackman:
The preferred commit log format is
`kubernetes-cli 1.0.0: Fix for Linuxbrew`

sjackman:
Thanks for your contribution, Andrew.

PR #491, opened 2015-07-22
vim: Add option to build vim with gtk2 gui

paulhybryant:
Add option to build vim with gtk2 gui.

### DIFF
--- a/Library/Formula/vim.rb
+++ b/Library/Formula/vim.rb
@@ -33,6 +33,7 @@ class Vim < Formula
   depends_on "luajit" => :optional
   depends_on :x11 if build.with? "client-server"
   depends_on "homebrew/dupes/ncurses" unless OS.mac?
+  depends_on "gtk+" => :optional
 
   conflicts_with "ex-vi",
     :because => "vim and ex-vi both install bin/ex and bin/view"
@@ -63,7 +64,11 @@ class Vim < Formula
     end
 
     opts << "--disable-nls" if build.include? "disable-nls"
-    opts << "--enable-gui=no"
+    if build.with? "gtk+"
+      opts << "--enable-gui=gtk2"
+    else
+      opts << "--enable-gui=no"
+    end
 
     if build.with? "client-server"
       opts << "--with-x"


Library/Formula/vim.rb
```diff
@@ -33,6 +34,7 @@ class Vim < Formula
   depends_on "luajit" => :optional
   depends_on :x11 if build.with? "client-server"
   depends_on "homebrew/dupes/ncurses" unless OS.mac?
+  depends_on "gtk+" if build.with? "gui-gtk2"
```
sjackman: Please change this line to `depends_on "gtk+" => :optional`

Library/Formula/vim.rb
```diff
@@ -15,6 +15,7 @@ class Vim < Formula
   option "override-system-vi", "Override system vi"
   option "disable-nls", "Build vim without National Language Support (translated messages, keymaps)"
   option "with-client-server", "Enable client/server mode"
+  option "with-gui-gtk2", "Build vim with GTK2 GUI"
```
sjackman: Please delete this line (see below).

Library/Formula/vim.rb
```diff
@@ -63,7 +65,11 @@ def install
     end
 
     opts << "--disable-nls" if build.include? "disable-nls"
-    opts << "--enable-gui=no"
+    if build.with? "gui-gtk2"
```
sjackman: Please change this line to `if build.with? "gtk+"`

sjackman:
Do you have access to a Mac OS machine? If this patch works on a Mac, it should be sent to Homebrew/homebrew rather than Homebrew/linuxbrew.

paulhybryant:
I have updated the pull request.
This does not work on Mac OS because some files are not available in the gtk+ installation on Mac OS (gdkx.h to be specific).

sjackman:
Merged. The preferred commit log format is:
`vim: Add option to build with gtk+ for Linuxbrew`

sjackman:
Thanks for your contribution, Paul.

PR #496, opened 2015-07-27
cmake depends on python on linux

dennyabraham:
I'm installing rust using linuxbrew on a new system that does not have python. It fails to build cmake since cmake is not bottled for this system and its python build dependency does not apply to linux. I added an explicit python dependency on linux. 

Will this respect a system-installed python as satisfying the dependency?

### DIFF
--- a/Library/Formula/cmake.rb
+++ b/Library/Formula/cmake.rb
@@ -31,6 +31,7 @@ class Cmake < Formula
   option "without-docs", "Don't build man pages"
 
   depends_on :python => :build if OS.mac? && MacOS.version <= :snow_leopard && build.with?("docs")
+  depends_on :python => :build if OS.linux?
   depends_on "curl" unless OS.mac?
 
   # The `with-qt` GUI option was removed due to circular dependencies if


sjackman:
Hi, Denny. This topic has been debated before. See #269. It was decided that many (most?) systems include a working `python`, and that `python` is a heavy-weight dependency to require just for `cmake`. You can install `cmake` without `python` like so:
`brew install cmake --without-docs`

dennyabraham:
Thanks! Are you still interested in a change to [python_requirement.rb](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/python_requirement.rb)?

sjackman:
Sure! What doesn't it do right now that it should do?

dennyabraham:
Nevermind, I misread #269. I was hoping the logic in python_requirement.rb was structured to check against the system-installed version.

sjackman:
It uses the `python` found in `PATH`, which can be `/usr/bin/python`:
https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/python_requirement.rb#L37

PR #499, opened 2015-07-30
New formula: iotop

tomas:
Second submission, this time `audit --strict` says we're A-OK. 

Given this is a python-based lib I might have missed something. I followed the instructions on the [wiki](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md), though.

### DIFF
--- a/Library/Formula/iotop.rb
+++ b/Library/Formula/iotop.rb
@@ -1,0 +1,22 @@
+class Iotop < Formula
+  desc "Top-like UI used to show which process is using the I/O"
+  homepage "http://guichaz.free.fr/iotop/"
+  url "http://guichaz.free.fr/iotop/files/iotop-0.6.tar.bz2"
+  sha256 "3adea2a24eda49bbbaeb4e6ed2042355b441dbd7161e883067a02bfc8dcef75b"
+  head "git://repo.or.cz/iotop.git"
+
+  depends_on :python
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"sbin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    out = shell_output("#{bin}/iotop --help")
+    assert_match "DISK READ and DISK WRITE", out
+  end
+end


tomas:
I should also note that iotop [requires](http://guichaz.free.fr/iotop/) a kernel above 2.6, and a system that provides a `/proc/self/io`. Is there any way brew can check these things before trying to install?

Library/Formula/iotop.rb
```diff
@@ -0,0 +1,22 @@
+class Iotop < Formula
+  desc "Top-like UI used to show which process is using the I/O"
+  homepage "http://guichaz.free.fr/iotop/"
+  url "http://guichaz.free.fr/iotop/files/iotop-0.6.tar.bz2"
+  sha256 "3adea2a24eda49bbbaeb4e6ed2042355b441dbd7161e883067a02bfc8dcef75b"
+  head "git://repo.or.cz/iotop.git"
+
+  depends_on "python"
```
sjackman: Use `depends_on :python` instead, which can use `/usr/bin/python`, whereas `"python"` must use a Homebrewed `python`.

Library/Formula/iotop.rb
```diff
@@ -0,0 +1,22 @@
...
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"sbin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    out = `iotop --help`
```
sjackman: Use `shell_output`. You'll find it in other formulas for examples.

tomas:
Ok, fixed.

Library/Formula/iotop.rb
```diff
@@ -0,0 +1,22 @@
...
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"sbin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    out = shell_output("#{bin}/iotop --help")
+    assert_match "DISK READ and DISK WRITE", out
```
sjackman: You can combine these into a single line. No need for the `out` variable.

sjackman:
Merged. Thanks for your contribution, Tomás. The preferred commit log message is
`iotop 0.6 (new formula) for Linuxbrew`

iggyvolz:
Just curious - why is this being merged straight into Linuxbrew?  Is it a Linux-specific command?

tomas:
I think I'm also curious by your question @iggyvolz. Isn't linuxbrew meant for Linux?

tomas:
Oh, I just realized what you mean. I'm not sure if iotop actually runs on OSX, but the documentation is pretty clear that it requires a *Linux* 2.6.x kernel or above. Is there a `/proc/self/io` path on OSX?

sjackman:
No, there's no `/proc/*` on Mac OS. It looks like Mac OS ships with an implementation of `iotop`, which is news to me:
```
~ ❯❯❯ which iotop
/usr/bin/iotop
~ ❯❯❯ iotop
dtrace: failed to initialize dtrace: DTrace requires additional privileges
```

iggyvolz:
@tomas Linuxbrew is a fork of Homebrew, which is designed for Mac.  Usually, new formulae (formulas?) are sent in to Homebrew, and then they're merged in here weekly - with minor changes for Linux compatibility if needed (hence all the "unless io.Mac?")

tomas:
@iggyvolz Yeah, I do use "regular" Homebrew on an old OSX laptop. I didn't know Homebrew formulae was ported to Linuxbrew on a weekly basis, though. 

Where should I submit a formula for something like GIMP or Inkscape?

sjackman:
If it compiles on both Mac and Linux, then to Homebrew/homebrew.

tomas:
Cool. Can you point to an example formula for a X11/GTK-based app?

dunn:
Both Inkscape and GIMP are already in the [X11 tap](https://github.com/Homebrew/homebrew-x11), actually!  But Inkscape might soon be moved to the core Homebrew/homebrew repository, since it doesn't depend on X11 anymore.

PR #501, opened 2015-08-03
emacs: add libxml2 compile option

danielschranz:
This was mostly pattern matching to the other options and a bit reading up on formulae as this is the first time I have touched one of those. 

Sorry if I didn't get it right, I'm happy to fix if it needs fixing - but it works fine for me and I can now use eww in emacs.

### DIFF
--- a/Library/Formula/emacs.rb
+++ b/Library/Formula/emacs.rb
@@ -26,6 +26,7 @@ class Emacs < Formula
 
   option "with-cocoa", "Build a Cocoa version of emacs"
   option "with-ctags", "Don't remove the ctags executable that emacs provides"
+  option "without-libxml2", "Don't build with libxml2 support"
 
   deprecated_option "cocoa" => "with-cocoa"
   deprecated_option "keep-ctags" => "with-ctags"
@@ -39,6 +40,7 @@ class Emacs < Formula
   depends_on "imagemagick" => :optional
   depends_on "mailutils" => :optional
   depends_on "glib" => :optional
+  depends_on "libxml2" unless OS.mac?
 
   # https://github.com/Homebrew/homebrew/issues/37803
   if build.with? "x11"
@@ -57,6 +59,12 @@ class Emacs < Formula
             "--infodir=#{info}/emacs"]
 
     args << "--with-file-notification=gfile" if build.with? "glib"
+
+    if build.with? "libxml2"
+      args << "--with-xml2"
+    else
+      args << "--without-xml2"
+    end
 
     if build.with? "d-bus"
       args << "--with-dbus"


iggyvolz:
If you made a PR on homebrew, it will get merged into Linuxbrew on a weekly basis.  Things only need to be committed directly to Linuxbrew if it's a Linux-specific fix, it's a Linux-specific software, or it's a critical security fix (in that case it's best to create an issue and ping Shaun so he merges manually)

danielschranz:
Thanks for the comment, I'll be sure to keep it in mind for next time.
I created a PR on homebrew for this (https://github.com/Homebrew/homebrew/pull/42429) - therefore closing this PR again.

mikemcquaid:
This is a Linux-specific fix because `libxml2` is installed by default on OS X. I'd rather if Linuxbrew didn't push these things upstream unless you're sure they're misplaced in Linuxbrew; I'm maybe wrong but it feels like lately there's been more PRs being directed to us that shouldn't be.

danielschranz:
To be fair the option itself and making use of the configure flag is generic to both projects, right? 

Only the dependency part might be specific to the OS and I think even that probably isn't necessary on Linux; at least on Linux Mint I had libxml2 already installed and I just put it in to make sure the build never breaks because of a missing libxml2 (a result of me not knowing enough about what we expect the user to have installed and what will be provided by other formulas)  

But if my PR on homebrew gets merged we will see if linuxbrew users will have issues with this flag on certain distributions and if this formula needs fixing here

sjackman:
There are 30 formulae in Homebrew that depend on `libxml2` so this PR didn't strike me as unusual.
```
❯❯❯ grep -l 'depends_on.*libxml2' * |column
aqbanking.rb            libdap.rb               nghttp2.rb
augeas.rb               libglade.rb             nordugrid-arc.rb
cfengine.rb             libimobiledevice.rb     openslide.rb
dnsperf.rb              libplist.rb             openwsman.rb
gdal.rb                 libslax.rb              postgres-xc.rb
gedit.rb                libspatialite.rb        postgresql.rb
gnome-doc-utils.rb      libvirt.rb              spdylay.rb
gtk-doc.rb              libxml++.rb             swish-e.rb
gwyddion.rb             libxmlsec1.rb           yelp-tools.rb
itstool.rb              libxslt.rb              yelp-xsl.rb
```

sjackman:
Six state that it's because the version of libxml2 in Mac OS version xxx is too old. The others give no explanation.
```
❯❯❯ grep 'depends_on.*libxml2' * |grep MacOS
cfengine.rb:  depends_on "libxml2" if MacOS.version < :mountain_lion
libxml++.rb:  depends_on "libxml2" if MacOS.version <= :leopard
libxmlsec1.rb:  depends_on "libxml2" if MacOS.version <= :lion
nghttp2.rb:  depends_on "libxml2" if MacOS.version <= :lion
postgres-xc.rb:  depends_on "libxml2" if MacOS.version <= :leopard # Leopard libxml is too old
postgresql.rb:  depends_on "libxml2" if MacOS.version <= :leopard # Leopard libxml is too old
```

sjackman:
My suggested fix for Linuxbrew would be
```ruby
depends_on "libxml2" unless OS.mac?
```
If you'd like to send up a PR with this content after testing that it works for you, I'll merge it in.

danielschranz:
Thanks, I'll try to update the PR in homebrew with that code so that we don't need to maintain a custom version for linuxbrew (I see that `Library/Formula/pypy.rb` is already using `OS.mac?` so I hope they won't object)

danielschranz:
As you've probably seen on the homebrew PR it has been suggested to have a separate PR to linuxbrew for 
```ruby 
depends_on "libxml2" unless OS.mac?
```

So the question now is - on MacOS libxml2 we've enabled support for `libxml2` by default  (changing the optiont to `--without-libxml2`) - do we want the same for linuxbrew (only difference being that the brew version of libxml2 will be used) or are we defaulting to only use libxml2 when the user specifies it (`--with-libxml2`)?

I am going to wait until the upstream changes arrived in linuxbrew and fix the libxml2 dependency then unless people want me to reopen this PR.

sjackman:
The `libxml2` dependency should be required on Linux.

danielschranz:
Re-opened the PR to add the libxml2 dependency

sjackman:
Hi, Daniel. Please squash and rebase this commit, and don't modify the bottle section.

danielschranz:
Sorry this took me a few days, hope it's correct now.

Library/Formula/emacs.rb
```diff
@@ -26,6 +26,7 @@ class Emacs < Formula
 
   option "with-cocoa", "Build a Cocoa version of emacs"
   option "with-ctags", "Don't remove the ctags executable that emacs provides"
+  option "without-libxml2", "Don't build with libxml2 support"
```
sjackman: Please remove this option.

Library/Formula/emacs.rb
```diff
@@ -58,6 +60,12 @@ def install
 
     args << "--with-file-notification=gfile" if build.with? "glib"
 
+    if build.with? "libxml2"
+      args << "--with-xml2"
+    else
+      args << "--without-xml2"
+    end
+
```
sjackman: Please remove this chunk as well.

sjackman:
I was able to build `emacs` without `libxml2`. What error message are you seeing when building `emacs`? Please report `brew gist-logs emacs`

PR #511, opened 2015-08-13
Implement a :blas dependency

tdsmith:
Example output with no system BLAS:

```
vagrant@vagrant-ubuntu-trusty-64:~$ brew install blastest
Warning: BLAS not configured
Falling back to brewed openblas. If you prefer to use a system BLAS,
please set HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS to correct
values.
Warning: BLAS not configured
Falling back to brewed openblas. If you prefer to use a system BLAS,
please set HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS to correct
values.
==> Downloading https://gist.githubusercontent.com/tdsmith/345b08bb656ad54c1701/raw/c0d4e2b1362d43dad0760bcc1ec24e4b881683ef/blastest.c
Already downloaded: /home/vagrant/.cache/Homebrew/blastest-1.0.c
==> /usr/bin/gcc-4.8 blastest.c -I/home/vagrant/.linuxbrew/opt/openblas/include -o blastest -L/home/vagrant/.linuxbrew/opt/openblas/lib -lopenb
/home/vagrant/.linuxbrew/Cellar/blastest/1.0: 2 files, 24K, built in 2 seconds

vagrant@vagrant-ubuntu-trusty-64:~$ ldd /home/vagrant/.linuxbrew/Cellar/blastest/1.0/bin/blastest
	linux-vdso.so.1 =>  (0x00007ffcee7a7000)
	libopenblas.so.0 => /home/vagrant/.linuxbrew/lib/libopenblas.so.0 (0x00007fdac8038000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fdac7c6c000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fdac7965000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fdac7747000)
	libgfortran.so.3 => /usr/lib/x86_64-linux-gnu/libgfortran.so.3 (0x00007fdac742d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fdac8eb7000)
	libquadmath.so.0 => /usr/lib/x86_64-linux-gnu/libquadmath.so.0 (0x00007fdac71f0000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fdac6fda000)
```

With a system BLAS:
```
vagrant@vagrant-ubuntu-trusty-64:~$ sudo apt-get install libblas-dev liblapack-dev
vagrant@vagrant-ubuntu-trusty-64:~$ brew reinstall blastest
==> Reinstalling blastest
==> Downloading https://gist.githubusercontent.com/tdsmith/345b08bb656ad54c1701/raw/c0d4e2b1362d43dad0760bcc1ec24e4b881683ef/blastest.c
Already downloaded: /home/vagrant/.cache/Homebrew/blastest-1.0.c
==> /usr/bin/gcc-4.8 blastest.c  -o blastest -lblas -llapack
/home/vagrant/.linuxbrew/Cellar/blastest/1.0: 2 files, 24K, built in 2 seconds

vagrant@vagrant-ubuntu-trusty-64:~$ ldd /home/vagrant/.linuxbrew/Cellar/blastest/1.0/bin/blastest
	linux-vdso.so.1 =>  (0x00007fff3cbfd000)
	libblas.so.3 => /usr/lib/libblas.so.3 (0x00007f0653dc9000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0653a04000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f06536fd000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f065404f000)
```

Setting HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS propagates:

```
vagrant@vagrant-ubuntu-trusty-64:~$ HOMEBREW_BLAS_CFLAGS="-I/tmp" HOMEBREW_BLAS_LDFLAGS="-lblas -llapack -lm" brew reinstall blastest
==> Reinstalling blastest
==> Downloading https://gist.githubusercontent.com/tdsmith/345b08bb656ad54c1701/raw/c0d4e2b1362d43dad0760bcc1ec24e4b881683ef/blastest.c
Already downloaded: /home/vagrant/.cache/Homebrew/blastest-1.0.c
==> /usr/bin/gcc-4.8 blastest.c -I/tmp -o blastest -lblas -llapack -lm
/home/vagrant/.linuxbrew/Cellar/blastest/1.0: 2 files, 24K, built in 2 seconds
```

This needs documentation (where?) and the warning message could probably be friendlier. It gets printed a couple times but I don't think that's easy to influence.

### DIFF
--- a/Library/Formula/blastest.rb
+++ b/Library/Formula/blastest.rb
@@ -1,0 +1,18 @@
+class Blastest < Formula
+  desc "blas test"
+  homepage "https://tds.xyz"
+  url "https://gist.githubusercontent.com/tdsmith/345b08bb656ad54c1701/raw/c0d4e2b1362d43dad0760bcc1ec24e4b881683ef/blastest.c"
+  sha256 "ea6a083d6bccbc94e503f00911e9abedb5e1b3a0ed80ad1e5077b80fd6f36b42"
+  version "1.0"
+
+  depends_on :blas if OS.linux?
+
+  def install
+    system "#{ENV.cc} blastest.c #{ENV["HOMEBREW_BLAS_CFLAGS"]} -o blastest #{ENV["HOMEBREW_BLAS_LDFLAGS"]}"
+    bin.install "blastest"
+  end
+
+  test do
+    system bin/"blastest"
+  end
+end
--- a/Library/Homebrew/requirements/blas_requirement.rb
+++ b/Library/Homebrew/requirements/blas_requirement.rb
@@ -1,0 +1,47 @@
+require "requirement"
+
+class BlasRequirement < Requirement
+  fatal true
+  default_formula "openblas"
+
+  # This ensures that HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS
+  # are always set. It does _not_ add them to CFLAGS or LDFLAGS; that
+  # should happen inside the formula.
+  env do
+    if @satisfied_result
+      ENV["HOMEBREW_BLAS_CFLAGS"] ||= ""
+      ENV["HOMEBREW_BLAS_LDFLAGS"] ||= "-lblas -llapack"
+    else
+      ENV["HOMEBREW_BLAS_CFLAGS"] = "-I#{Formula["openblas"].opt_include}"
+      ENV["HOMEBREW_BLAS_LDFLAGS"] = "-L#{Formula["openblas"].opt_lib} -lopenblas"
+    end
+  end
+
+  satisfy :build_env => true do
+    cflags = ENV["HOMEBREW_BLAS_CFLAGS"] || ""
+    ldflags = ENV["HOMEBREW_BLAS_LDFLAGS"] || "-lblas -llapack"
+    success = nil
+    Dir.mktmpdir do |tmpdir|
+      tmpdir = Pathname.new tmpdir
+      (tmpdir/"blastest.c").write <<-EOS.undent
+        double cblas_ddot(const int, const double*, const int, const double*, const int);
+        int main() {
+          double x[] = {1.0, 2.0, 3.0}, y[] = {4.0, 5.0, 6.0};
+          cblas_ddot(3, x, 1, y, 1);
+          return 0;
+        }
+      EOS
+      success = system "#{ENV["CC"]} #{cflags} #{tmpdir}/blastest.c -o #{tmpdir}/blastest #{ldflags}",
+                :err => "/dev/null"
+    end
+    if !success
+      opoo "BLAS not configured"
+      puts <<-EOS.undent
+        Falling back to brewed openblas. If you prefer to use a system BLAS,
+        please set HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS to correct
+        values.
+      EOS
+    end
+    success
+  end
+end
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -126,6 +126,8 @@ class DependencyCollector
       Dependency.new("libtool", tags)
     when :python2
       PythonRequirement.new(tags)
+    when :blas
+      BlasRequirement.new(tags)
     else
       raise ArgumentError, "Unsupported special dependency #{spec.inspect}"
     end
--- a/Library/Homebrew/requirements.rb
+++ b/Library/Homebrew/requirements.rb
@@ -1,5 +1,6 @@
 require "requirement"
 require "requirements/apr_requirement"
+require "requirements/blas_requirement"
 require "requirements/fortran_requirement"
 require "requirements/language_module_requirement"
 require "requirements/minimum_macos_requirement"


tdsmith:
Also, do we need separate CFLAGS and FFLAGS?

Library/Formula/blastest.rb
```diff
@@ -0,0 +1,18 @@
+class Blastest < Formula
+  desc "blas test"
+  homepage "https://tds.xyz"
+  url "https://gist.githubusercontent.com/tdsmith/345b08bb656ad54c1701/raw/c0d4e2b1362d43dad0760bcc1ec24e4b881683ef/blastest.c"
+  sha256 "ea6a083d6bccbc94e503f00911e9abedb5e1b3a0ed80ad1e5077b80fd6f36b42"
+  version "1.0"
+
+  depends_on :blas if OS.linux?
```
dpo: I don't think that `if OS.linux?` is necessary since `libblas` and `liblapack` are symlinked to `Accelerate` on osx.

dpo:
+1

davydden:
thanks for working on it!

with respect to different `CFLAGS` and `FFLAGS`, i am not sure. @Rombur Bruno, any ideas from your experience installing on clusters with candi?

Library/Homebrew/requirements/blas_requirement.rb
```diff
@@ -0,0 +1,47 @@
+require "requirement"
+
+class BlasRequirement < Requirement
+  fatal true
+  default_formula "openblas"
+
+  # This ensures that HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS
+  # are always set. It does _not_ add them to CFLAGS or LDFLAGS; that
+  # should happen inside the formula.
```
davydden: you asked about documentation, not sure if that's what you mean, but perhaps one could quickly summarise the logic here, i.e. : to check the presence of system's blas we compile a test program with HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS if they are set or use the default values "" and "-lblas -llapack". If compilation fails we fall back to openblas and set HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS accordingly. Otherwise, user specified or default values are stored.

Rombur:
I don't think it is necessary to have different ```CFLAGS``` and ```FFLAGS``` I can't think of a situation where they would need to be different.

davydden:
any updates on this PR? @tdsmith have you decided if it can be merged upstream or is it bound to be in `linuxrew` only?

tdsmith:
This will be Linuxbrew-only. I was going to let Shaun merge this but I seem to remember he's on vacation.

dpo:
I thought `:blas` could also apply to OSX.

davydden:
+1 to what @dpo said, i would really like to see it upstream.

@tdsmith would you be willing to reconsider the decision? Let me explain how introducing `:blas` in upstream could help: 

1)  If we keep :blas in Linuxbrew only each and every formula in `Homebrew-science` which needs blas/lapack would need to have something like
```
depends_on :blas if os.liinux?
ENV["HOMEBREW_BLAS_LDFLAGS"] = "-lblas -llapack" if os.mac?
```
and then we use those `HOMEBREW_BLAS_CFLAGS` (will be empty on OSX) and `HOMEBREW_BLAS_LDFLAGS`. That should work, but is less elegant than simply having 
```
depends_on :blas
```
One could make the latter case work even if OSX `Homebrew` would have a super simple `blas_requirement.rb` which just sets those variables to the default values:
```
require "requirement"

class BlasRequirement < Requirement
      ENV["HOMEBREW_BLAS_CFLAGS"] = ""
      ENV["HOMEBREW_BLAS_LDFLAGS"] = "-framework Accelerate"
end
```
so that we always try to build against `-framwrok Accelerate`. 

In those rare cases when we need `-velibfort` on OS-X, we'll have to have something like
```
ENV["HOMEBREW_BLAS_LDFLAGS"] = "-lveclibfort" if os.mac?
```

2) If :blas is in OS-X `Homebrew`, that means that every formula within Homebrew does not need to have a special treatment to work on Linuxbrew fork as far as `blas/lapack` is concerned. That's  saves modifications/commits/maintainance efforts in the Linuxbrew project.

Personally i don't see how Homebrew itself would suffer from this decision. In many formula `-framework Accelerated` is already hardcoded. We gradually change it once to `HOMEBREW_BLAS_LDFLAGS` and it's done.

Irrespectively of whether `:blas` goes upstream or not, I am very glad that it'll be in Linuxbrew as this should allow to use the latter for high-perfomance-computing. So I very welcome this PR in any case.

davydden:
p.s. if on os-x you would like to allow users build against `openblas`, then there could be a simple environment flag on OSX only `HOMEBREW_BLAS=openblas` which shall be set at the very beginning (like optional HOMEBREW_GCC=xxx). And then inside `blas_requirement.rb` you would decide on whether require `openblas` or use `-framework Accelerate`. That in itself will allow to cut out a number of code duplication inside Homebrew formulae. So that shifts the management of blas/lapack from each formula to that `blas_requirement.rb`. 

I think it's win-win-win for `Homebrew`, `Linuxbrew` and `Homebrew-science`.

Rombur:
I just wanted to add that, since today, MKL is free (https://software.intel.com/en-us/articles/free_mkl). This will probably increased the number of people using different version of BLAS/LAPACK on OSX.

tdsmith:
Oh, that's very interesting. If MKL is freely available that makes me think maybe there is real value in making this flexible on OS X.

tdsmith:
MKL does not actually appear to be available in the community license package yet? https://software.intel.com/en-us/forums/topic/590743 -- https://www.dropbox.com/s/mgtd2cncyqmbpbq/Screenshot%202015-08-31%2014.29.39.png?dl=0 vs https://www.dropbox.com/s/28rkxwlsuo4niin/Screenshot%202015-08-31%2014.30.00.png?dl=0

tdsmith:
Intel confirmed in that thread that MKL is not available for free as a standalone product on OS X.

tjhei:
I played around with this PR and I would like to give some feedback:
- I think blas/lapack should be considered together.
- There is a large number of compatible system blas/lapack libraries: openblas, former gotoblas, fblaslapack, mkl, atlas, ...  and it would be great to detect any of them (automatically? here is a list of libs that we check in deal.II: [1])
- Even if you find blas, not everything might be exposed and I am not sure how to handle that (here is the check we use in deal.II: [2])
- Ideally, the additional code would end up in a formula so that it can be modified by tapping a repository and not in the main linuxbrew code. I wonder how difficult it is to introduce a meta package that does not contain any own code, but exports blas/lapack dirs either from a system library or a formula that has been installed.

[1]: https://github.com/dealii/dealii/blob/98dc9319a5a3e6ddff7e65360ebb57eaec8ef836/cmake/modules/FindLAPACK.cmake
[2]: https://github.com/dealii/dealii/blob/e29135bb547281dacccd4802f6a864af0c80dbdd/cmake/configure/configure_1_lapack.cmake

tdsmith:
Do you know what platforms on which you see incomplete BLASses?

Why would a tap want to affect BLAS selection? I can imagine a way you'd do this in a formula but it'd be super-weird. Handling this per-tap is more complexity than I think is warranted but maybe I don't understand the use case.

tjhei:
> Do you know on which platforms you see incomplete BLASses?

I don't, sorry. I assume this is also to make sure we get the BLAS and LAPACK symbols (they are sometimes in separate libs and sometimes in a single lib).

> Why would a tap want to affect BLAS selection? I can imagine a way you'd do this in a formula but it'd be super-weird. Handling this per-tap is more complexity than I think is warranted but maybe I don't understand the use case.

I think there is some value in having the option to develop, improve, test, and easily distribute these meta formulas. Otherwise one would need to distribute a modified linuxbrew itself.

tdsmith:
To answer an earlier point: I agree blas and lapack should be considered together and this PR does that in practice. Is there a name or symbol that's clearer than `blas` that conveys this? I think `linalg` is vague and `blas_and_or_lapack` is clunky.

davydden:
how about `:blaslapack`  as it's BLAS+LAPACK.

dpo:
I like `:blaslapack`, but perhaps `:blas` is sufficient if we need to keep it short. For example, OpenBLAS contains both BLAS and LAPACK subroutines.

davydden:
> I think there is some value in having the option to develop, improve, test, and easily distribute these meta formulas. Otherwise one would need to distribute a modified linuxbrew itself.

how many of those meta-formulas do you envision to have @tjhei ? If there are very few (like `:cmake` for which I created a PR), i would vote to have extra symbols which support those **essential for compilation of third-party-software** system libraries (`:mpi`, `:blaslapack`, `:cmake`) to be inside `linuxbrew` repo. And then if soemthing does not work on a particular platform, it would be matter of fixing it in a fork of linuxbrew and creating a PR, i guess.

tjhei:
> how many of those meta-formulas do you envision to have @tjhei ?

Not sure. I am still trying to understand linuxbrew and its philosophy and I am probably trying to do wrong things in the process. :-) The idea that different formulas can provide same/similar functionality and a way to express this using dependencies sounds like a reasonable thing to have for a package manager.
Another thing I don't understand is how you guys deal with different versions of libraries. It seems that there is version support in dependencies, but old versions in formulas seem to be replaced instead of added.

tjhei:
I think the next thing where I will get annoyed is compiling python from scratch.

davydden:
> I think the next thing where I will get annoyed is compiling python from scratch.

hopefully should not happen as there is https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/requirements/python_requirement.rb

tdsmith:
BLAS_LAPACK / :blas_lapack seems reasonable and is clearer.

The mechanism for dependencies which may be satisfied from more than one package or from the environment is a Requirement. This PR provides an example of a Requirement implementation. They live in core. The [homebrew-php tap](https://github.com/Homebrew/homebrew-php) has engineered a way to carry Requirements in a tap, but probably does not allow overriding already-defined Requirements in core, which is probably fine.

We mostly do not deal with multiple versions of libraries; the most recent release is supported and others are not unless a formula exists for it in homebrew-versions.

davydden:
> BLAS_LAPACK / :blas_lapack seems reasonable and is clearer.

sounds good to me.

tdsmith:
I'll submit a PR to Homebrew this weekend; I think the case for getting rid of all the optional openblas dependencies in homebrew-science is pretty compelling and maybe we'll see MKL on OS X someday.

davydden:
@tdsmith any update on `:blas_lapack` in Homebrew?

davydden:
i was playing with your branch now. It seems we would need more functionality: for example, Hypre configures with
```
./configure \
  --prefix=LIBRARY-INSTALL-LOCATION \
  --with-MPI \
  --with-examples \
  --with-blas-libs="mkl_core mkl_intel_lp64 mkl_sequential" \
  --with-blas-lib-dirs=/opt/intel/mkl/11.1.064/lib/em64t \ 
  --with-lapack-libs="mkl_core mkl_intel_lp64 mkl_sequential" \
  --with-lapack-lib-dirs=/opt/intel/mkl/11.1.064/lib/em64t \
```
so i would say there should be:

1. a set of libraries names (`mkl_core`,`mkl_intel_lpi64`,`mkl_sequential`)
2. lib dir (`${MKLROOT}/lib/intel64`)
3. include dir (`${MKLROOT}/include`)

these three, as far as i can tell, should be enough to construct linker flags (e.g. `-L${MKLROOT}/lib/intel64 -lmkl_gf_lp64 -lmkl_core -lmkl_sequential`), compiler flags (e.g. `-I${MKLROOT}/include`) and also make it configurable with `--with-blas-libs=xxx`.

davydden:
on a bright side, by using 

1. current `:blas` dependency (from this PR) 
2. some hardcoding as mentioned in the above
3. symlinking cmake to avoid its installation via linuxbrew
4. compiling ruby (very simple, just `./configure --prefix=path/to/`)

I was able to install `petsc`, `slepc`, `trilinos`, `p4est`, `hypre` on a CentOS cluster with native `openmpi`, `gcc`, `mkl`, `cmake` and `git`.

So the bottom line -- it is **certainly worth having :blas_lapack**, just need to have more flexible API for blas/lapack to support all those different usage cases in configuring other libraries.

p.s. the only extra/auxilary packages i had to install are `curl`, `makedepend`, `openssl`, `pkg-config`, `zlib`.

davydden:
ps. updated the list of libraries i tried

davydden:
I updated the API and have a fully working (tested on CentOS with MKL) :blas dependency to install the mentioned in the above formulae. I will create a PR to Homebrew tomorrow...

davydden:
@tdsmith i updated your code in https://github.com/Homebrew/linuxbrew/pull/592

ahundt:
@tdsmith there are actually several implementations of blas so configuration of blas would absolutely make sense on OS X. There is a large number of blas implementations which would be useful across multiple platforms such as: [nvidia cublas](https://developer.nvidia.com/cublas), [lapack](http://www.netlib.org/lapack/), [intel mkl](http://software.intel.com/en-us/intel-mkl/), [amd acml](http://developer.amd.com/tools-and-sdks/cpu-development/amd-core-math-library-acml/), [netlib blas](http://www.netlib.org/blas/) (netlib comes in both c and fortran flavors).

Also please note that blas doesn't always require fortran.

Note that there seems to be a new approach happening at https://github.com/Homebrew/homebrew-science/pull/2889, it would be nice it would also work when fortran isn't required.

davydden:
@ahundt that's the same approach I am trying to push to `homebrew-science`, the difference is that requirement is local to the tap and not global in Homebrew/Linuxbrew ;-) To see that in action you can check https://github.com/davydden/homebrew-dealiisuite

dpo:
@ahundt I don't think ACML is relevant here because Apple computers don't have AMD processors.

ahundt:
@dpo fair enough (if we pretend linuxbrew doesn't exist). Anyway, I think the original point stands because I was just illustrating the variety of blas packages rather than accurately enumerating them.

sjackman:
Should this issue be moved to Homebrew/science?

davydden:
it's there for a while https://github.com/Homebrew/homebrew-science/pull/2889

sjackman:
Great. Thanks, Denis.

PR #519, opened 2015-08-24
libcroco: Needs libxml2 on Linux

dholm:


Library/Formula/libcroco.rb
```diff
@@ -14,6 +14,7 @@ class Libcroco < Formula
   depends_on "pkg-config" => :build
   depends_on "intltool" => :build
   depends_on "glib"
+  depends_on "libxml2" if OS.linux?
```
danabrand: Thanks for your contribution. Please change this to
```ruby
depends_on "libxml2" unless OS.mac?
```

dholm: Fixed!

sjackman:
Fixes the error: `No package 'libxml-2.0' found`

sjackman:
Thanks for your contribution, David.

PR #520, opened 2015-08-24
elfutils 0.163

dholm:


sjackman:
Do you have access to a Mac, and does `elfutils` compile on a Mac? If so, submit it upstream to https://github.com/Homebrew/homebrew

dholm:
I do.  I'll see if it builds.

dholm:
@sjackman `configure` fails with Clang due to missing support for the GNU extension [Nested Functions](https://gcc.gnu.org/onlinedocs/gcc/Nested-Functions.html):

    configure: error: gcc with GNU99 support required

sjackman:
Add to the formula:
```ruby
fails_with :clang do
  build 700
  cause "gcc with GNU99 support required"
end
```

dholm:
@sjackman Done.

sjackman:
Does it build on Mac now?

Library/Formula/elfutils.rb
```diff
@@ -0,0 +1,29 @@
+class Elfutils < Formula
+  desc "Libraries and utilities for handling ELF objects."
+  homepage "https://fedorahosted.org/elfutils/"
+  url "https://fedorahosted.org/releases/e/l/elfutils/0.163/elfutils-0.163.tar.bz2"
+  sha256 "7c774f1eef329309f3b05e730bdac50013155d437518a2ec0e24871d312f2e23"
+
+  depends_on "bzip2"
+  depends_on "xz"
+  depends_on "zlib"
```
sjackman: Please change to
```ruby
  depends_on "xz"
  depends_on "bzip2" unless OS.mac?
  depends_on "zlib" unless OS.mac?
```

sjackman:
The build fails with GCC 5.3.0:
```
checking for __thread support... no
configure: error: __thread support required
```

sjackman:
Merged. Thanks for your contribution, David. Sorry for the extended delay.

sjackman:
Just out of curiosity, what do you use `elfutils` for instead of `binutils`?

dholm:
@sjackman elfutils provides libraries that are used by many other applications, such as [perf](https://perf.wiki.kernel.org/index.php/Main_Page).

PR #521, opened 2015-08-24
python: Fix Linux build

dholm:
Some of the search paths are OSX-specific and should not be modified on
Linux.  The OpenSSL helper needs to be enabled or the SSL modules won't
be built.


tdsmith:
This should be redundant with #460.

dholm:
I've tried #460 and it seems to solve all the issues that I had.  I'm closing this pull request.

PR #522, opened 2015-08-25
Tor formula fixed for linux

nstCactus:
Fixes the tor formula for Linux (libevent wasn't found by the `./configure` script).
Tested on a Raspberry Pi 2 and a Macbook running OS X 10.10.5 beta.


sjackman:
Can you report the exact error that you see without this patch?

sjackman:
I wasn't able to reproduce this issue on standalone Linuxbrew.

sjackman:
Is this issue still a problem for you?

PR #543, opened 2015-09-08
ghc 7.10.2: Fix for linuxbrew

paulhybryant:
Fix the error:
```
error while loading shared libraries: libgmp.so.3: cannot open shared object file: No such file or directory
```

### DIFF
--- a/Library/Formula/ghc.rb
+++ b/Library/Formula/ghc.rb
@@ -63,10 +63,11 @@ class Ghc < Formula
     # GMP *does not* use PIC by default without shared libs  so --with-pic
     # is mandatory or else you'll get "illegal text relocs" errors.
     resource("gmp").stage do
-      system "./configure", "--prefix=#{gmp}", "--with-pic", "--disable-shared"
+      system "./configure", "--prefix=#{gmp}", "--with-pic"
       system "make"
       system "make", "check"
       ENV.deparallelize { system "make", "install" }
+      ln_s Dir["#{gmp}/lib/libgmp.so.*"], "#{gmp}/lib/libgmp.so.3"
     end
 
     args = ["--with-gmp-includes=#{gmp}/include",
@@ -82,6 +83,7 @@ class Ghc < Formula
       args << "--with-gcc-4.2=#{ENV.cc}"
     end
 
+    ENV.prepend_path "LD_LIBRARY_PATH", "#{gmp}/lib"
     resource("binary").stage do
       # Change the dynamic linker and RPATH of the binary executables.
       if OS.linux? && Formula["glibc"].installed?


sjackman:
I didn't find this patch was necessary to compile ghc 7.10.2 for me. I had previously created a symlink in `$HOMEBREW_PREFIX/lib` from `libgmp.so.10.2.0` to `libgmp.so.3`. Perhaps that's why?

Library/Formula/ghc.rb
```diff
@@ -63,10 +63,11 @@ def install
     # GMP *does not* use PIC by default without shared libs  so --with-pic
     # is mandatory or else you'll get "illegal text relocs" errors.
     resource("gmp").stage do
-      system "./configure", "--prefix=#{gmp}", "--with-pic", "--disable-shared"
+      system "./configure", "--prefix=#{gmp}", "--with-pic"
       system "make"
       system "make", "check"
       ENV.deparallelize { system "make", "install" }
+      system "ln", "-s", "#{gmp}/lib/libgmp.so.10.2.0", "#{gmp}/lib/libgmp.so.3"
```
dunn: Homebrew has a method for installing symlinks:
```ruby
(gmp/"lib").install_symlink "#{gmp}/lib/libgmp.so.10.2.0", "libgmp.so.3"
```

paulhybryant:
Yes, the old symlink is why this patch is not needed.

Library/Formula/ghc.rb
paulhybryant: Thanks, I have updated the PR accordingly.

sjackman:
I'd be curious to know why linking statically against `gmp` doesn't work on Linux, as it seems to work on Mac.

paulhybryant:
Linking statically against gmp works fine on linux.
It is the step after making ghc fails because of missing shared library

==> Downloading http://downloads.haskell.org/~ghc/7.6.3/ghc-7.6.3-x86_64-unknown-linux.tar.bz2
==> ./configure --prefix=/tmp/ghc20151102-21968-1q007is/ghc-7.10.2/binary --with-gmp-includes=/home/paulhybryant/.linuxbrew/Cellar/ghc/7.10.2/libexec/integer-gmp/include --with-gmp-libraries=/home/paulhybryant/.linuxbrew/Cellar/ghc/7.10.2/libexec/integer-gmp
Last 15 lines from /home/paulhybryant/.cache/Homebrew/Logs/ghc/05.configure:
2015-11-02 16:49:02 +0800

./configure
--prefix=/tmp/ghc20151102-21968-1q007is/ghc-7.10.2/binary
--with-gmp-includes=/home/paulhybryant/.linuxbrew/Cellar/ghc/7.10.2/libexec/integer-gmp/include
--with-gmp-libraries=/home/paulhybryant/.linuxbrew/Cellar/ghc/7.10.2/libexec/integer-gmp/lib
--with-ld=ld
--with-gcc=/home/paulhybryant/.linuxbrew/bin/ygcc-5

configure: WARNING: unrecognized options: --with-ld
checking for path to top of build tree... utils/ghc-pwd/dist-install/build/tmp/ghc-pwd: error while loading shared libraries: libgmp.so.3: cannot open shared object file: No such file or directory
configure: error: cannot determine current directory
/home/paulhybryant/.linuxbrew/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: ./configure --prefix=/tmp/ghc20151102-21968-1q007is/ghc-7.10.2/binary --with-gmp-includes=/home/paulhybryant/.linuxbrew/Cellar/ghc/7.10.2/libexec/integer-gmp/include --with-gmp-libraries=/home/paulhybryant/.linuxbrew/Cellar/ghc/7.10.2/libexec/integer-gmp/lib --with-ld=ld --with-gcc=/home/paulhybryant/.linuxbrew/bin/gcc-5

sjackman:
I'm on vacation until the new year. Please bump this thread after then if you haven't heard from me.

paulhybryant:
This PR no longer works for some reason. Now it complains about not being able to link to gmp

sjackman:
I just pushed ghc 7.10.3b. Any luck with that?

paulhybryant:
ghc install would still fail without the change here, complaining about not being able to find libgmp.so.3

With the change in this PR it now works.

Thanks,

paulhybryant:
And the change for linking the dynamic library should be
ln_sf "#{gmp}/lib/libgmp.so.10.3.0", "#{gmp}/lib/libgmp.so.3"

I will update the PR.

sjackman:
How about
```ruby
ln_s Dir["#{gmp}/lib/libgmp.so.*"], "#{gmp}/lib/libgmp.so.3"
```

paulhybryant:
Thanks, I am not very familiar with ruby.
What does Dir["#{gmp}/lib/libgmp.so.*"] mean? Which file will libgmp.so.3 point to?

sjackman:
`Dir[]` is the globbing operator, just like shell globbing. http://ruby-doc.org/core-2.2.3/Dir.html#method-c-5B-5D

paulhybryant:
Thanks, updated the PR.

sjackman:
This patch would have the side effect of making ghc dynamically linked against libgmp, which is different than it's done on Mac. I'm hesitant to make this change for this reason.

To run the command `ghc-pwd` we need a gmp shared library named `libgmp.so.3` somewhere in the library search path. We can get the shared library by installing the formula `gmp`, so I suggest adding the following line instead of removing `"--disable-shared"`.

```ruby
depends_on "gmp" => :build unless OS.mac?
```
In `def install`
```ruby
if OS.linux?
  # Fix error while loading shared libraries: libgmp.so.3
  ln_s Formula["gmp"].lib/"libgmp.so", "#{gmp}/lib/libgmp.so.3"
  ENV.prepend_path "LD_LIBRARY_PATH", "#{gmp}/lib"
end
```

An alternative solution would be for the `gmp` formula to provide the symlink from `libgmp.so.10` to `libgmp.so.3`.

sjackman:
Can you confirm that the dependency on `libgmp.so.3` is only a build time dependency? That is, none of the executables installed by `ghc` depend on it?

sjackman:
Fixed. Thanks for your contribution, Paul. Sorry for the extended delay.

sjackman:
The fix that I went with is:
```diff
diff --git a/Library/Formula/ghc.rb b/Library/Formula/ghc.rb
index 2eec23b..b01e24b 100644
--- a/Library/Formula/ghc.rb
+++ b/Library/Formula/ghc.rb
@@ -14,6 +14,8 @@ class Ghc < Formula
   deprecated_option "tests" => "with-test"
   deprecated_option "with-tests" => "with-test"
 
+  depends_on "gmp" => :build if OS.linux?
+
   resource "gmp" do
     url "http://ftpmirror.gnu.org/gmp/gmp-6.1.0.tar.bz2"
     mirror "https://gmplib.org/download/gmp/gmp-6.1.0.tar.bz2"
@@ -81,14 +83,23 @@ def install
       args << "--with-gcc-4.2=#{ENV.cc}"
     end
 
+    if OS.linux?
+      # Fix error while loading shared libraries: libgmp.so.3
+      ln_s Formula["gmp"].lib/"libgmp.so", gmp/"lib/libgmp.so.3"
+      ENV.prepend_path "LD_LIBRARY_PATH", gmp/"lib"
+      # Fix /usr/bin/ld: cannot find -lgmp
+      ENV.prepend_path "LIBRARY_PATH", gmp/"lib"
+    end
+
     resource("binary").stage do
       # Change the dynamic linker and RPATH of the binary executables.
       if OS.linux? && Formula["glibc"].installed?
```

paulhybryant:
Thank you for the fix and maintaining linuxbrew!

sjackman:
No worries. I'm glad that it's useful.

PR #544, opened 2015-09-10
Fixed 404 error from oniguruma

romejoe:
Prior url gave 404 error when trying to install. Updated formula with the latest homebrew version


dunn:
Changes in Homebrew are merged into Linuxbrew about once a week, so Homebrew/homebrew@ 5eed615a12a62de9fa6bea599cb9f28bbef1eb0b will appear shortly!

PR #547, opened 2015-09-14
updated to gradle 2.7

gcstang:
new gradle 2.7 released upgraded to 2.7


sjackman:
New versions should be sent to Homebrew/hombrew. See https://github.com/Homebrew/linuxbrew/blob/master/CONTRIBUTING.md

Issue #534, opened 2015-09-03
Can't install vim


dky:
Hello, I can't install vim. Looks like the download URL is pointing to a invalid location:

$ brew install vim --with-lua
==> Downloading https://mirrors.kernel.org/debian/pool/main/v/vim/vim_7.4.712.orig.tar.gz

curl: (22) The requested URL returned error: 404 Not Found
Error: Failed to download resource "vim"
Download failed: https://mirrors.kernel.org/debian/pool/main/v/vim/vim_7.4.712.orig.tar.gz

dky:
Edit to resolve I used "brew edit vim" and updated the url to something that existed, ran md5 check on file and updated values. All good.

dunn:
The main Homebrew repo has an updated vim, so once it gets merged into Linuxbrew this will be fixed.

destroyerofbuilds:
@dky thank you for filing this issue, and documenting the resolution. I ended up running into the same issue recently while reinstalling my LinuxBrew setup, and your suggestion gave me a start on resolving it locally.

zhe:
I had the same 404 problem a while ago. After `brew doctor` and `brew update`, Homebrew is updated from `ce6d82b7` to `d430de47`. Then, I run
```
brew install vim --with-lua
```
It works without error now. 
Homebrew downloads Vim from https://github.com/vim/vim/archive/v7.4.889.tar.gz

PR #550, opened 2015-09-15
Fix broken vim package, upgrade vim to 7.4.826

outcoldman:
vim_7.4.712.orig.tar.gz does not exist on the server


destroyerofbuilds:
@outcoldman please see #534 

We should probably just merge in upstream from Homebrew instead of fixing this directly in LinuxBrew.

sjackman:

PR #551, opened 2015-09-15
Fix cv formula

outcoldman:
Current sha is mismatched for version 0.7.1.
Version 0.9 is called now progress instead of cv.


dunn:
This was done upstream in https://github.com/Homebrew/homebrew/commit/c02e6dcb92941356c7d0d8f8e07b927f1f8aeb01; it'll be merged into Linuxbrew soon.  If you want to update the Homebrew formula from 0.8 to 0.9, that'd be great!

sjackman:

PR #552, opened 2015-09-15
Fix Oniguruma formula

outcoldman:
The old url returns 403. Copied formula from homebrew repo.

I copy pasted formula from https://github.com/Homebrew/homebrew/blob/c47b806f5a593cac262c10b72a936e86f90b3027/Library/Formula/oniguruma.rb


sjackman:
Homebrew is merged into Linuxbrew periodically. See https://github.com/Homebrew/linuxbrew/blob/master/CONTRIBUTING.md

PR #553, opened 2015-09-15
Update jdk to upstream version 1.8.0-60

brevilo:


brevilo:
I couldn't really test it locally but the URL and the sha256sum are correct...

sjackman:
The git log message should be simply `jdk 1.8.0-60`

sjackman:
Thanks for your contribution, Oliver.

brevilo:
The git log message should be simply jdk 1.8.0-60

Next time :smile:

PR #555, opened 2015-09-16
Implement a :cmake dependency

davydden:
the rationale is to avoid installing (and possibly failing while doing so) a big number of dependencies when `cmake` is available on the target system. Essentially similar to `:mpi`.

~~This is work in progress (inspired by `:emacs` dependency). I certainly parse versions correctly but i think something is wrong in the comparision as if a test formula needs cmake `2.7` and `2.8.12.2` is available, the linuxbrew `cmake` build is still triggered...~~

related PR: https://github.com/Homebrew/linuxbrew/pull/511
partly addresses the issue: https://github.com/Homebrew/linuxbrew/issues/504

### DIFF
--- a/Library/Homebrew/requirements/cmake_requirement.rb
+++ b/Library/Homebrew/requirements/cmake_requirement.rb
@@ -1,0 +1,36 @@
+require "requirement"
+
+class CmakeRequirement < Requirement
+  fatal true
+  default_formula "cmake"
+
+  def initialize(tags)
+    @version = tags.shift if /\d+\.*\d*/ === tags.first
+    raise "Specify a version for CmakeRequirement" unless @version
+    super
+  end
+
+  satisfy :build_env => false do
+    next unless which "cmake"
+    cmake_version = Utils.popen_read("cmake", "--version")
+    words = cmake_version.split(" ")
+    cmake_version = words.last
+    # debug:
+    ohai "cmake_version: #{cmake_version} vs #{@version}"
+    Version.new(cmake_version) >= Version.new(@version)
+  end
+
+  env do
+    ENV.prepend_path "PATH", which("cmake").dirname
+  end
+
+  def message
+    s = "Cmake #{@version} or later is required."
+    s += super
+    s
+  end
+
+  def inspect
+    "#<#{self.class.name}: #{name.inspect} #{tags.inspect} version=#{@version.inspect}>"
+  end
+end
--- a/Library/Homebrew/requirements/cmake_requirement.rb
+++ b/Library/Homebrew/requirements/cmake_requirement.rb
@@ -12,11 +12,7 @@ class CmakeRequirement < Requirement
 
   satisfy :build_env => false do
     next unless which "cmake"
-    cmake_version = Utils.popen_read("cmake", "--version")
-    words = cmake_version.split(" ")
-    cmake_version = words.last
-    # debug:
-    ohai "cmake_version: #{cmake_version} vs #{@version}"
+    cmake_version = Utils.popen_read("cmake", "--version").split(" ").last
     Version.new(cmake_version) >= Version.new(@version)
   end
 
--- a/Library/Homebrew/requirements/cmake_requirement.rb
+++ b/Library/Homebrew/requirements/cmake_requirement.rb
@@ -17,7 +17,9 @@ class CmakeRequirement < Requirement
   end
 
   env do
-    ENV.prepend_path "PATH", which("cmake").dirname
+    # do append_path to avoid messing up with isolated build
+    # envirnoment when cmake is located in /usr/bin/ and alike
+    ENV.append_path "PATH", which("cmake").dirname
   end
 
   def message
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -114,6 +114,7 @@ class DependencyCollector
     when :ant        then ant_dep(spec, tags)
     when :apr        then AprRequirement.new(tags)
     when :emacs      then EmacsRequirement.new(tags)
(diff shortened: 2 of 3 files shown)


davydden:
it is actually working! :+1  i got internal cmake triggered from dependencies of the package where i still had `depends_on "cmake" => :build` instead of `depends_on :cmake => ["2.8", :build]`

davydden:
would certainly like it to be upstream, but @tdsmith won't agree :-)

Library/Homebrew/requirements/cmake_requirement.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  satisfy :build_env => false do
+    next unless which "cmake"
+    cmake_version = Utils.popen_read("cmake", "--version").split(" ").last
+    Version.new(cmake_version) >= Version.new(@version)
+  end
+
+  env do
+    ENV.prepend_path "PATH", which("cmake").dirname
```
tjhei: Not 100% sure, but if my cmake binary sits in say /usr/bin/, this will pull that into the path possibly messing with the isolated build environment or not?

tdsmith:
At least I'm predictable! Homebrew wouldn't contemplate this because cmake doesn't exist on the OS X base system and the number of potential failures and the challenge of debugging them increase sharply if Homebrew doesn't control the dependency tree.

I think Linuxbrew should reject this because the cmake build isn't very painful and I want to avoid hearing about problems with whatever ancient cmake they ship with CentOS 6 but I'm trying very hard not to become a Linuxbrew maintainer so my opinion probably doesn't matter. ;)

davydden:
> I think Linuxbrew should reject this because the cmake build isn't very painful and I want to avoid hearing about problems with whatever ancient cmake they ship with CentOS 6 but I'm trying very hard not to become a Linuxbrew maintainer so my opinion probably doesn't matter. ;)

ain't gonna be problems with ancient cmake as we have minimum versions here. As to painfulness, i think @tjhei can share his concerns and issue with not using system's cmake.

tjhei:
> can share his concerns and issue with not using system's cmake.

This is my experience (minimal ubuntu):
```
$ brew install cmake
==> Installing dependencies for cmake: makedepend, zlib, openssl, pkg-config, curl
==> Installing cmake dependency: makedepend
Error: /home/bob/dealii-brew/linuxbrew/opt/pkg-config not present or broken
Please reinstall pkg-config. Sorry :(

$ brew info cmake
==> Dependencies
Required: curl ✘

$ brew info curl
==> Dependencies
Build: pkg-config ✘
Required: openssl ✘

$ brew info pkg-config
pkg-config: stable 0.28
Manage compile and link flags for libraries
https://wiki.freedesktop.org/www/Software/pkg-config/
Not installed
From: https://github.com//homebrew/blob/master/Library/Formula/pkg-config.rb

$ brew install pkg-config
# ok, this worked?!

$ brew install cmake
==> Installing cmake dependency: makedepend
==> Installing cmake dependency: zlib
==> Installing cmake dependency: openssl
==> Installing cmake dependency: curl
==> Installing cmake
# fails (no python installed)

$ sudo apt-get install python (I don't want to brew install python!!)

$ brew install cmake
ImportError: No module named setuptools

$ sudo apt-get install python-setuptools

$ brew install cmake
# finally worked.
```

tdsmith:
I do see two bona fide Linuxbrew bugs here:
* pkg-config isn't installed automatically
* Homebrew assumes that there is a system Python with a functioning setuptools so Linuxbrew assumes that as well; it's either a Linuxbrew documentation error that this isn't specified or a Linuxbrew error that the formula doesn't spec a :python dependency.

I don't think using a system cmake helps.

davydden:
i agree that there are some errors in the cmake related formulas which should be addressed, but I also believe that `linuxbrew` should not force users to install `cmake` (and its dependencies) if system cmake is available and we can do without. Why would one spend more time doing it if there is a shortcut?.. 

Checking out what is the minimum cmake version required for a particular software is super simple, it is one of the first lines in `CMakeList.txt` : `CMAKE_MINIMUM_REQUIRED(VERSION 2.8.8)`. Thus all the hassles with old system cmake can be avoided. And if we are not sure what's the minimum version (i.e. developers did not specify it), we can still use `depends_on "cmake" => :build ` instead of `depends_on :cmake => ["2.8.8", :build]`.

All in all i do not see any downsides of this PR. It only increase the flexibility of linuxbrew thus making it more attractive for users which value using system provided **basic building libraries** (cmake, mpi, blas-lapack, gcc, gfortran).

Let's see what  @sjackman has to say when he's back from holidays...

Rombur:
I just found this PR and I just want to say that I disagree with @tdsmith about the fact that installing cmake is not painful. I can go on cmake website and get the binary for linux. This is as fast as it can be. However, with linuxbrew, I have to compile openssl! This is painfully slow. 

If you don't like the idea of using cmake from the system, what about using curl from the system? This would speed up installation a lot and I doubt that many people require a specific version of curl.

sjackman:
@tdsmith wrote…
> I'm trying very hard not to become a Linuxbrew maintainer

Come over to the dark side!

sjackman:
I prefer to avoid the system libraries and binaries as much as possible and use Homebrew versions. The various wacky old versions of system tools give me many headaches. If you really want to use the system `cmake` the workaround is:
```
mkdir -p `brew --prefix`/Cellar/cmake/3.3.2/bin
ln -s /usr/bin/cmake `brew --prefix`/Cellar/cmake/3.3.2/bin/
```
Untested, but you get the gist of it.

I generally won't accept features like this into Linuxbrew that won't ever make it into Homebrew. I try to keep the diff between Homebrew and Linuxbrew as small as possible to make maintenance and merging of Homebrew/Linuxbrew as easy as possible.

PR #556, opened 2015-09-16
ace: Corrected homepage url, upgraded to latest 6.3.2, removed mentioning …

jwillemsen:
…of configure, that has been removed from ACE years ago


dunn:
Changes that aren't specific to Linuxbrew are made upstream and marged in about once a week. Once https://github.com/Homebrew/homebrew/pull/44020 is merged, we'll get the fix here too!

sjackman:

PR #558, opened 2015-09-17
qt (add to caveat)

davydden:
without those packages:
```
Basic XLib functionality test failed!
 You might need to modify the include and library search paths by editing
 QMAKE_INCDIR_X11 and QMAKE_LIBDIR_X11 in /home/davydden/Library/Temp/qt20150917-1700-1na8nib/qt-everywhere-opensource-src-4.8.7/mkspecs/linux-g++.
```


sjackman:
For the X11 packages that do not exist in Linuxbrew, I'd prefer to move add a subsection to the `README.md` [Dependencies](https://github.com/Homebrew/linuxbrew#dependencies) for installing X11 dependencies on Debian and Fedora systems.

PR #559, opened 2015-09-17
freeimage (3.17.0)

davydden:
fix for gcc-5 and corrected patches in makefiles.

### DIFF
--- a/Library/Formula/freeimage.rb
+++ b/Library/Formula/freeimage.rb
@@ -24,7 +24,15 @@ class Freeimage < Formula
 
   option :universal
 
-  patch :DATA
+  patch :DATA if OS.mac?
+
+  # same as the :DATA patch, but for linux
+  if OS.linux?
+    patch do
+      url "https://gist.githubusercontent.com/davydden/5a4f348108d3c9110299/raw/3396840ff71c639d848ce552de9124462777ab97/freeimage.patch"
+      sha256 "537a4045d31a3ce1c3bab2736d17b979543758cf2081e97fff4d72786f1830dc"
+    end
+  end
 
   # fix GCC 5.0 compile.
   patch do


sjackman:
This should be sent upstream to Homebrew if it addresses compiler errors with GCC 5 on Mac as well as Linux.

davydden:
it's not only that! it's also changes of `.dylib` to `.so` in patches. So the one you have in Linuxbrew is not working irrespectively of gcc version.

davydden:
i submitted gcc related stuff here: https://github.com/Homebrew/homebrew/pull/44612

sjackman:
This patch is too big for me to merge into Linuxbrew. My recommendation would be to pull the inline patch out into gist and send what portion of this patch upstream to Homebrew/homebrew that isn't specific to Linux. Send the remaining portion that's specific to Linux here to Homebrew/linuxbrew.

davydden:
i already send the gcc-related part to homebrew (see link above). Once it's merged/resolved there, i will update the PR here.

sjackman:
Great. Thanks, Denis. Blocked on Homebrew/homebrew#44612

davydden:
fixed in upstream. After you merge it I will rebase this PR.

sjackman:
Thanks, Denis. The merge will come in the new year.

sjackman:
I've merged Homebrew into Linuxbrew.

davydden:
i've rebased

sjackman:
This patch is too large for Linuxbrew.
Does it work to simply change
```ruby
  patch :DATA
```
to
```ruby
  patch :DATA if OS.mac?
```

davydden:
are you sure you are looking at the right version? The patch is roughly the same number of lines as the one in Homebrew. I just modified it to work on Linux.

sjackman:
Yep. [Showing  1 changed file  with 20 additions and 56 deletions.](https://github.com/Homebrew/linuxbrew/pull/559/files)

Does it work if you simply disable this patch as above?

davydden:
i don't need to disable it to see that it will not work otherwise ;-)
seriously, just see what it does to Makefile which grants a working installation.

sjackman:
That whole patch appears to be Mac specific. Try disabling it.

davydden:
[these](https://github.com/davydden/linuxbrew/blob/freeimage_linuxbrew/Library/Formula/freeimage.rb#L97-L101) and [these](https://github.com/davydden/linuxbrew/blob/freeimage_linuxbrew/Library/Formula/freeimage.rb#L109-L110) and [these](https://github.com/davydden/linuxbrew/blob/freeimage_linuxbrew/Library/Formula/freeimage.rb#L118-L123) lines in the patch have nothing to do with MAC and as I said they grant working installation. 
The mac-version is similar in that sense.

davydden:
to close this topic once and forever, ignoring the patch gives:
```
Last 15 lines from /home/davydden/Library/Logs/Homebrew/freeimage/02.make:
2016-01-07 10:52:19 +0100

make
-f
Makefile.gnu
install
PREFIX=/home/davydden/.linuxbrew/Cellar/freeimage/3.17.0

install -d //usr/include //usr/lib
install -m 644 -o root -g root Source/FreeImage.h //usr/include
install: cannot create regular file ‘//usr/include/FreeImage.h’: Permission denied
make: *** [install] Error 1
```

sjackman:
Okay. Have you sent this patch to the upstream author? https://github.com/davydden/linuxbrew/blob/freeimage_linuxbrew/Library/Formula/freeimage.rb#L45-L127

davydden:
upstream is already fine. I have not changed this part of the patch in the upstream, it was working without any issues. On the contrary, I **adjusted** this patch to work for Linux. That's what the whole PR is about...

sjackman:
I wasn't clear enough. I meant have you sent this patch upstream to the freeimage author.

I don't want to merge this patch as it is now for two reasons. One, it breaks the build on Mac OS and Linuxbrew aims to be compatible with both Mac and Linux. Two, if Homebrew changes the patch for Mac it will create merge conflicts when I merge Homebrew into Linuxbrew.

I suggest gist-ing your patch for Linux, and patching this formula like so:
```ruby
if OS.mac?
  patch :DATA
else
  patch do
    …
  end
end
```

davydden:
it does not have to be sent upstream to freeimage. It mostly settings for makefile, it does not fix any bugs.

davydden:
I will shift it to the gist later....

davydden:
@sjackman i shifted it to the gist

sjackman:
Thanks for your contribution, Davy. For future contributions, the preferred commit log message is
`freeimage 3.17.0: Fix for Linuxbrew`

davydden:
@sjackman I am Denis :smile:

sjackman:
:blush: My apologies, Denis.

PR #565, opened 2015-09-22
boost: Add the Homebrew library path in linkflags

tueda:
In my case 9a98daf (cf. #322) was not enough, and I got linker errors during making shared libraries (libboost_iostreams-mt.so.1.58.0 and libboost_iostreams.so.1.58.0):

```
/usr/bin/ld: cannot find -lbz2
```

This patch makes b2 program use the correct Homebrew library path.


sjackman:
Thanks for your contribution, Takahiro.

PR #569, opened 2015-09-25
Bumped Solr to 5.3.1

brevilo:


sjackman:
New versions should be sent upstream to https://github.com/Homebrew/homebrew

PR #570, opened 2015-09-26
Fix Go on Linux.

DoomHammer:
Pretty much what it says. Use Linux resources instead of Darwin ones.

### DIFF
--- a/Library/Formula/go.rb
+++ b/Library/Formula/go.rb
@@ -28,9 +28,12 @@ class Go < Formula
     if MacOS.version > :lion
       url "https://storage.googleapis.com/golang/go1.4.2.darwin-amd64-osx10.8.tar.gz"
       sha256 "c2f53983fc8fe5159d811081022ebc401b8111759ce008f91193abdae82cdbc9"
-    else
+    elsif OS.mac?
       url "https://storage.googleapis.com/golang/go1.4.2.darwin-amd64-osx10.6.tar.gz"
       sha256 "da40e85a2c9bda9d2c29755c8b57b8d5932440ba466ca366c2a667697a62da4c"
+    else
+      url "https://storage.googleapis.com/golang/go1.4.2.linux-amd64.tar.gz"
+      sha256 "141b8345932641483c2437bdbd65488a269282ac85f91170805c273f03dd223b"
     end
   end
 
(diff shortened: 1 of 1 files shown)


Library/Formula/go.rb
```diff
@@ -49,7 +52,7 @@ def install
 
     cd "src" do
       ENV["GOROOT_FINAL"] = libexec
-      ENV["GOOS"]         = "darwin"
+      ENV["GOOS"]         = OS.mac? ? "darwin" : "linux"
```
sjackman:
```ruby]
ENV["GOOS"]         = os = OS.mac? ? "darwin" : "linux"
```

Library/Formula/go.rb
```diff
@@ -76,7 +79,11 @@ def install
         cd "src/golang.org/x/tools/cmd/vet/" do
           system "go", "build"
           # This is where Go puts vet natively; not in the bin.
-          (libexec/"pkg/tool/darwin_amd64/").install "vet"
+          if OS.mac?
+            (libexec/"pkg/tool/darwin_amd64/").install "vet"
+          else
+            (libexec/"pkg/tool/linux_amd64/").install "vet"
+          end
```
sjackman: Please use instead
```ruby
(libexec/"pkg/tool/#{os}_amd64/").install "vet"
```

Library/Formula/go.rb
```diff
@@ -112,8 +119,13 @@ def caveats; <<-EOS.undent
...
     if build.with? "vet"
-      assert File.exist?(libexec/"pkg/tool/darwin_amd64/vet")
-      assert File.executable?(libexec/"pkg/tool/darwin_amd64/vet")
+      if OS.mac?
+        assert File.exist?(libexec/"pkg/tool/darwin_amd64/vet")
+        assert File.executable?(libexec/"pkg/tool/darwin_amd64/vet")
+      else
+        assert File.exist?(libexec/"pkg/tool/linux_amd64/vet")
+        assert File.executable?(libexec/"pkg/tool/linux_amd64/vet")
+      end
```
sjackman: Ditto, use `#{os}` as above.

DoomHammer:
All suggested fixes in place.

Library/Formula/go.rb
```diff
@@ -49,7 +52,7 @@ def install
 
     cd "src" do
       ENV["GOROOT_FINAL"] = libexec
-      ENV["GOOS"]         = "darwin"
+      ENV["GOOS"]         = os
```
sjackman: I think you're missing the following line?
```ruby
os = OS.mac? ? "darwin" : "linux"
```

DoomHammer:
You are right. Fixed it.

xu-cheng:
@DoomHammer Thanks for the pull request! 🎉

sjackman:
Thanks for merging, Xu. I appreciate your help with Linuxbrew.

Issue #579, opened 2015-10-03
Can't install PMD


rwhogg:
Installing PMD on Ubuntu 14.04 fails with the following error:
```
tar -xf /tmp/pmd20151002-6066-1wef7se/pmd-src-5.3.3/pmd-dist/target/pmd-bin-5.3.3.zip
tar: This does not look like a tar archive
tar: Skipping to next header
tar: Exiting with failure status due to previous errors
/home/bob/.linuxbrew/Library/Homebrew/debrew.rb:11:in `raise'
ErrorDuringExecution: Failure while executing: tar -xf /tmp/pmd20151002-6066-1wef7se/pmd-src-5.3.3/pmd-dist/target/pmd-bin-5.3.3.zip
```

Here's the stack trace:

```
/home/bob/.linuxbrew/Library/Homebrew/debrew.rb:11:in `raise'
/home/bob/.linuxbrew/Library/Homebrew/utils.rb:225:in `safe_system'
/home/bob/.linuxbrew/Library/Formula/pmd.rb:15:in `install'
/home/bob/.linuxbrew/Library/Homebrew/debrew.rb:22:in `block in install'
/home/bob/.linuxbrew/Library/Homebrew/debrew.rb:97:in `debrew'
/home/bob/.linuxbrew/Library/Homebrew/debrew.rb:22:in `install'
/home/bob/.linuxbrew/Library/Homebrew/build.rb:129:in `block in install'
/home/bob/.linuxbrew/Library/Homebrew/formula.rb:848:in `block in brew'
/home/bob/.linuxbrew/Library/Homebrew/formula.rb:1389:in `block in stage'
/home/bob/.linuxbrew/Library/Homebrew/resource.rb:91:in `block in unpack'
/home/bob/.linuxbrew/Library/Homebrew/extend/fileutils.rb:17:in `mktemp'
/home/bob/.linuxbrew/Library/Homebrew/resource.rb:88:in `unpack'
/home/bob/.linuxbrew/Library/Homebrew/resource.rb:81:in `stage'
/usr/lib/ruby/1.9.1/forwardable.rb:201:in `stage'
/home/bob/.linuxbrew/Library/Homebrew/formula.rb:1381:in `stage'
/home/bob/.linuxbrew/Library/Homebrew/formula.rb:844:in `brew'
/home/bob/.linuxbrew/Library/Homebrew/build.rb:107:in `install'
[...]

PR #580, opened 2015-10-03
pmd: Use unzip instead of tar to extract on Linux

rwhogg:
The PMD formula previously used tar(1) to extract the zip archive. This works
fine on Macs, because BSD's tar can handle zip archives. GNU tar can't, so as
of this commit we use unzip(1) instead.

By the way, happy hacktoberfest! https://hacktoberfest.digitalocean.com/

### DIFF
--- a/Library/Formula/pmd.rb
+++ b/Library/Formula/pmd.rb
@@ -12,7 +12,10 @@ class Pmd < Formula
     doc.install "LICENSE", "NOTICE", "README.md"
 
     # The mvn package target produces a .zip with all the jars needed for PMD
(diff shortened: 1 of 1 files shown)


rwhogg:
Oh, forgot to say this: it fixes #579

sjackman:
@tdsmith Would you accept a patch upstream to Homebrew/homebrew that uses `unzip` rather than `tar` to extract a .zip file?

tdsmith:
Yep!

sjackman:
@rwhogg Please remove the `OS` conditionals and send this patch upstream to https://github.com/Homebrew/homebrew
Thanks for your contribution, Bob.

PR #582, opened 2015-10-05
[WIP] append to LD_LIBRARY_PATH formula.lib

davydden:
tries to fix https://github.com/Homebrew/homebrew-science/issues/2521

I do append to `LD_LIBRARY_PATH`, but netcdf still fails :(

What works **inside** Netcdf is
```
if build.with? "fortran"
  resource("fortran").stage do
        ENV.prepend "LD_LIBRARY_PATH", "#{lib}" # <=============
        system "./configure", *common_args
        system "make"
        system "make", "check" if build.with? "check"
        system "make", "install"
  end
end
```

could that be related to `resource()` usage?


davydden:
i close this as I don't have time looking into it.

Issue #528, opened 2015-09-01
Travis


tdsmith:
What's the story with Linuxbrew and Travis? Noticing that Travis supports Docker now we ought to be able to abuse that to test on Ubuntu and CentOS images.

tdsmith:
-- and the standalone image, which will make bottling easier.

tdsmith:
Docs here: http://docs.travis-ci.com/user/docker/

dpo:
We haven't had too much success so far. One blocking factor is the time it takes to build stuff (especially gcc and the like). We get kicked out.

ahundt:
could the same bottle integration the mac version has speed up some of those steps for things that depend on gcc?

saketkc:
@dpo did manage to hack around that. 
See: https://github.com/Homebrew/homebrew-science/issues/1795#issuecomment-75081443

xu-cheng:
See https://github.com/Homebrew/homebrew-science/pull/2847

sjackman:
I'd happily accept a pull request updating `.travis.yml` on Linuxbrew. I have limited time for troubleshooting, because I'm on parental leave. My son Jake is two months old. In one week I'll be leaving with my family on vacation until Christmas.

saketkc:
Here is a hacked version of `.travis.yml' building: https://travis-ci.org/saketkc/linuxbrew/builds/85045113

It currently complains of `git` missing, which should probably go to @sjackman's Dockerfile?

I am a novice docker user and I am not even sure, if it is going to do properly what it is supposed to.

DoomHammer:
Same here at https://travis-ci.org/DoomHammer/linuxbrew . Not sure about this `git` as it is installed. I guess it might have some library loading problem or maybe something's wrong with repository? Would give it further look later.

Strange thing is, it builds locally under similar Docker pipeline.

DoomHammer:
OK, some success. Now it fails locally, too :)

DoomHammer:
OK. [This one](https://travis-ci.org/DoomHammer/linuxbrew/jobs/85057681) works albeit with a tiny hack. I had to manually `apt-get install git` since _somewhere_ during `test-bot` run my linuxbrew repo is cleared and I have no software brewed whatsoever.

Anybody have any idea how to properly fix this instead of working around?

DoomHammer:
Opened #601 to discuss my approach there.

rwhogg:
[...]

PR #601, opened 2015-10-13
Use Docker in .travis.yml

DoomHammer:
First, slightly hackish approach. It should handle bottling for shorter builds at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 rvm:
   - 2.0.0
+services: docker
 
 before_install:
   # Fix error: Incorrect file permissions (664)
@@ -9,13 +10,25 @@ before_install:
   - if [ -f ".git/shallow" ]; then
       travis_retry git fetch --unshallow;
     fi
+  - export TRAVIS_COMMIT=$(git rev-parse --verify -q HEAD)
+  - docker pull linuxbrew/linuxbrew;
+    docker run
+      -v $(pwd):/home/linuxbrew/linuxbrew
+      linuxbrew/linuxbrew
+      /bin/sh -c 'brew update && sudo apt-get update && sudo apt-get install git -y && cp -rf ~/linuxbrew/* ~/linuxbrew/.git $(brew --repo) && brew fetch openssl && brew install ruby';
+    docker commit $(docker ps -l -q) linuxbrew;
+  - env | grep TRAVIS | tee /tmp/travis.env
 
 install:
   - PATH=$PWD/bin:$PATH
   - export HOMEBREW_DEVELOPER="1"
 
 script:
-  - brew test-bot
+  - docker run
+      --env-file /tmp/travis.env
+      -t
+      linuxbrew
+      brew test-bot;
 
 notifications:
   email:


dpo:
Is there a way to package Java in the docker images? Look at this "successful" build: https://travis-ci.org/Homebrew/homebrew-science/jobs/86294570.

sjackman:
Sure, Java can be installed with apt-get or brew. It'll have to wait until I'm back from vacation in the new year.

DoomHammer:
With all the talk about leaving Homebrew organization maybe it is a good time to revisit those changes as Linuxbrew would definitely benefit from any kind of working CI?

sjackman:
Yes, absolutely. Getting Travis CI up and running is my next project after Linuxbrew moves to its own organization.

DoomHammer:
I'll wait here, then :watch:

sjackman:
Travis CI for Linuxbrew is up and running! I haven't looked into Docker yet.

DoomHammer:
Let me know when you want to look at it, I'll rebase the changes then.

sjackman:
I had planned on building bottles on the `sjackman/linuxbrew-standalone` Docker image. I'm starting to think that there are some benefits to building the bottles on an older distribution. The compiled bottles should work on Linuxbrew Standalone, but should also work on older systems. The dependencies used from the host system should be `glibc` and `libstdc++`. All the other dependencies should be provided by Linuxbrew. That way perhaps Linuxbrew users could use bottles without having to install `glibc` and `gcc` (for `libstdc++`).

Linuxbrew: glibc 2.19

Ubuntu 14: glibc 2.19
Ubuntu 12: glibc 2.15

CentOS 7: glibc 2.17
CentOS 6: glibc 2.12
CentOS 5: glibc 2.5

DoomHammer:
So, do you suggest using Ubuntu 12 as a base for Docker/Travis instead of current `linuxbrew-standalone`?

sjackman:
I've been using `Ubuntu 14` as the base for Docker/Travis instead of `linxubrew-standalone`. It has the same version of `glibc` used by Linuxbrew (2.19), which is a nice coincidence. The bottles built on Ubuntu 14 work great on Linuxbrew. I've tested a few now.

sjackman:
Hi, Piotr. I've created a new [Dockerfile](https://github.com/Linuxbrew/linuxbrew/blob/master/Dockerfile) for Linuxbrew, and there's a new [Linuxbrew/linuxbrew](https://hub.docker.com/r/linuxbrew/linuxbrew/) docker image created from it.

sjackman:
`brew tests` is oddly failing right now.
```
1) Failure:
IntegrationCommandTests#test_uses [/home/linuxbrew/.linuxbrew/Library/Homebrew/test/test_integration_cmds.rb:629]:
--- expected
+++ actual
@@ -1,2 +1,2 @@
-"testball1
-testball2"
+"testball2
+testball1"
2) Failure:
LanguageModuleRequirementTests#test_good_python_deps [/home/linuxbrew/.linuxbrew/Library/Homebrew/test/test_language_module_requirement.rb:43]:
Expected #<LanguageModuleRequirement: "languagemodule" [:python, "datetime", nil]>
to be satisfied?.
```

DoomHammer:
I have uploaded a new version. I guess all those `if [ "${TRAVIS_OS_NAME}" = "osx" ]` are no longer necessary? Or do you plan to support OS X in the long term?

DoomHammer:
Also: I am not sure whether test-bot behaves correctly, since it tries to tap `doomhammer/linuxbrew` which in practice means `doomhammer/homebrew-linuxbrew` (which does not exist).

sjackman:
> Or do you plan to support OS X in the long term?

For the package manager `brew` and for for the formula, but not for the `.travis.yml`.

sjackman:
> Also: I am not sure whether test-bot behaves correctly, since it tries to tap doomhammer/linuxbrew which in practice means doomhammer/homebrew-linuxbrew (which does not exist).

That's a bug. Feel free to submit a pull request if you are able to fix it.

DoomHammer:
I am not familiar with `brew` internals. Could you point me what is the best way to start, since `test-bot.rb` is pretty big?

DoomHammer:
I believe I have found the problematic line, but I don't know the original authors idea. I believe @xu-cheng have added it.

Why is there `if user == "Homebrew" && repo == "homebrew"`? Doesn't it prevent Pull Requests from being tested (as they clearly have a different user, but the same repo name)?

Library/Homebrew/tap.rb
```diff
@@ -31,7 +31,7 @@ def self.fetch(*args)
     repo = repo.strip_prefix "homebrew-"
 
     if user == "Homebrew" && repo == "homebrew" ||
-        user == "Linuxbrew" && repo == "linuxbrew"
+        repo == "linuxbrew"
```
xu-cheng: This is incorrect. The core tap should always call `Homebrew/homebrew` even for Linuxbrew. Noted this is not the name of Linuxbrew project but  an internal name for a specialize tap.

DoomHammer: Yes, I believe this is incorrect, I only wanted to test it. In my case I get `user == "DoomHammer" && `repo == "linuxbrew"` when running `test-bot` on my PR on Travis. Could you help me find out what is wrong here?

xu-cheng: You need to update https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/cmd/test-bot.rb#L44
to
```ruby
if (tap = ENV["TRAVIS_REPO_SLUG"]) && (tap =~ %r{^([\w-]+)/(homebrew-)?([\w-]+)$})
```

xu-cheng: Sorry, the regex should be `%r{^([\w-]+)/homebrew-([\w-]+)$}`. I wrote it wrong in the first time.

DoomHammer:
After dealing with missing dependency on `rake` (by running `brew install ruby`) I have issues with `brew tests` as seen in [https://travis-ci.org/DoomHammer/linuxbrew/builds/114097361](Travis log). Any ideas what am I missing?

Most interestging lines are:

```
 1) Failure:
FormulaTests#test_eligible_kegs_for_cleanup [/home/linuxbrew/.linuxbrew/Library/Homebrew/test/test_formula.rb:365]:
--- expected
+++ actual
@@ -1,2 +1,2 @@
 [#<Keg:/tmp/homebrew_tests20160306-1572-kmxsq4/cellar/testball/0.2>,
- #<Keg:/tmp/homebrew_tests20160306-1572-kmxsq4/cellar/testball/0.3>]
+ #<Keg:/tmp/homebrew_tests20160306-1572-kmxsq4/cellar/testball/0.1>]
```

sjackman:
That's a weird failure. I'm not sure why the test is saying that 0.2 and 0.3 are expected to be cleaned up. Perhaps you could troubleshoot by printing out the value of `f3.installed_kegs` right before:
```ruby
    assert_equal f3.installed_kegs[0..1], f3.eligible_kegs_for_cleanup
```

DoomHammer:
Running this locally I get slightly different error, but:

```
KEGGIE: [#<Keg:/tmp/homebrew_tests20160307-182-ykgdsq/cellar/testball/0.3>, #<Keg:/tmp/homebrew_tests20160307-182-ykgdsq/cellar/testball/0.1>, #<Keg:/tmp/homebrew_tests20160307-1
82-ykgdsq/cellar/testball/0.2>]
```

then

```
  1) Failure:
FormulaTests#test_eligible_kegs_for_cleanup [/home/linuxbrew/.linuxbrew/Library/Homebrew/test/test_formula.rb:366]:
--- expected
+++ actual
@@ -1,2 +1,2 @@
-[#<Keg:/tmp/homebrew_tests20160307-182-ykgdsq/cellar/testball/0.3>,
- #<Keg:/tmp/homebrew_tests20160307-182-ykgdsq/cellar/testball/0.1>]
+[#<Keg:/tmp/homebrew_tests20160307-182-ykgdsq/cellar/testball/0.1>,
+ #<Keg:/tmp/homebrew_tests20160307-182-ykgdsq/cellar/testball/0.2>]
```

Also:

```
  2) Failure:
IntegrationCommandTests#test_uses [/home/linuxbrew/.linuxbrew/Library/Homebrew/test/test_integration_cmds.rb:629]:
--- expected
+++ actual
@@ -1,2 +1,2 @@
-"testball1
-testball2"
+"testball2
+testball1"
```

sjackman:
I've seen this test failure too. See #895

DoomHammer:
Now the #895 is gone, Travis CI is passing. I understand it is better to wait for [upstream](https://github.com/Homebrew/homebrew/commit/7970035e5d8bdaf201e44b0fa84a047f9f67ab22) to be merged rather than committing local changes? BTW: What is this ci/dockercloud thing?

sjackman:
I've merged Homebrew into Linuxbrew just now. I'm testing out using Docker Cloud running on Amazon EC2 as a CI for building bottles for Linuxbrew.
https://docs.docker.com/docker-cloud/feature-reference/automated-build/

DoomHammer:
Well, when I click "Details" I get 404 and I wanted to know what did I break (as it says "Your tests failed in Docker Cloud").

sjackman:
The link works for me. I'll guess that it's a permissions issue, that you don't have the perms to view the build log. Once it's up and running I'll sort that out, hopefully. Here's the log:
```
==> brew doctor
==> FAILED
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!
Warning: Suspicious git origin remote found.
With a non-standard origin, Homebrew won't pull updates from
the main repository. The current git origin is:
https://9a1bc74ab023883d39a8ba7b145913ebb40de5e1@github.com/DoomHammer/linuxbrew.git
Unless you have compelling reasons, consider setting the
origin remote to point at the main repository, located at:
https://github.com/Linuxbrew/linuxbrew.git
```
The error message is because `brew test-bot` isn't yet set up to integrate with Docker Cloud CI. I'm working on it.

sjackman:
What about the Travis failure though. Does that make sense to you?

DoomHammer:
Ah, so I plead not guilty, then for Docker CI. As for the Travis, it just timed out. I made it re-run the tests, but they are at the edge of timing-out too. Building `ruby` is the most time-consuming thing here, but couldn't do it properly without it. One possible idea would be to have Docker image prebuilt with `ruby`.

DoomHammer:
The little guy is green now :)

DoomHammer:
Sorry to be bothering you, @sjackman, but are we go with this?

sjackman:
> Building ruby is the most time-consuming thing here

Can we use the Ruby deb rather than brew ruby?

sjackman:
No apologies needed for bothering me. What are the advantages as you see it of using a Docker container rather than using the usual Travis build bot?

DoomHammer:
Ruby itself is not a problem. `brew` uses `rake` which is a separate package on Debian and somehow it didn't seem to be doing alright.

As for why Docker? Honestly, can't remember but this PR was a result of some discussion, IIRC. History says it was probably #528 .

One thing for certain: playing with Docker is more reproducible, so theoretically anybody can look at Travis error and then check it locally.

sjackman:
> One thing for certain: playing with Docker is more reproducible, so theoretically anybody can look at Travis error and then check it locally.

That is in advantage. It is quite frustrating trying to reproduce a Travis build failure locally. Do you know whether the Travis build machine images are available on Docker Hub?

DoomHammer:
No idea about the Travis build machines but in my PR I am making use of `linuxbrew/linuxbrew` (same thing I usually do prior to reporting bug, as this image I consider 'sane' contrary to most of machines I work on).

sjackman:
Hi, Piotr. I would like to continue testing Ubuntu 12 precise using Travis CI. It's important to test that installing `glibc` and `gcc` automatically works for older distributions.
Ubuntu 14 trusty is being tested using Docker cloud. That's my biggest reason for not using this patch. I also want to test Linuxbrew on a variety of platforms, and `linuxbrew/linuxbrew` is already being tested on Docker Cloud.
I know that this patch precedes my use of Docker Cloud, which is relatively recent.
I have thought about enabling Travis testing on both Ubuntu 12 and Ubuntu 14, but not necessarily using Docker. The downside to adding more dists to the testing matrix, is that tests will take longer, and it's nice to get quick feedback when iteratively testing updating a PR.

DoomHammer:
Makes sense to me. Closing this then?

sjackman:
Yes, I think so. Sorry, Piotr. This PR raised some helpful discussion. Thanks for your effort.

DoomHammer:
No problem. Not all ideas are good ones ;)

PR #603, opened 2015-10-13
glibc: enable obsolete rpc

DoomHammer:
Necessary to compile libvirt.

### DIFF
--- a/Library/Formula/glibc.rb
+++ b/Library/Formula/glibc.rb
@@ -16,6 +16,7 @@ class Glibc < Formula
         "--disable-debug",
         "--disable-dependency-tracking",
         "--disable-silent-rules",
+        "--enable-obsolete-rpc",
         "--prefix=#{prefix}",
         "--without-selinux"] # Fix error: selinux/selinux.h: No such file or directory
       args << "--with-binutils=" +


sjackman:
How is this handled in other distributions of Linux? I see that [Linux from Scratch](http://www.linuxfromscratch.org/lfs/view/development/chapter06/glibc.html) uses `--enable-obsolete-rpc`. Why is RPC obsolete, and what is supposed to replace it?

DoomHammer:
It was like that: tried installing libvirt but it failed being unable to find some headers (namely rpc/xdr.h and rpc/rpc.h). I looked up on packages.debian.org (as I usually do) and found them to belong to glibc. Its [ChangeLog](http://savannah.gnu.org/forum/forum.php?forum_id=7288) states the following:

>New configure option --enable-obsolete-rpc makes the deprecated RPC
>headers and functions available at compile time as they were before 
>version 2.14. This option will be removed at some time in the future 
>after the TI-RPC library becomes fully sufficient for the needs of 
>existing applications.

sjackman:
Okay, so TI-RPC is the replacement, but it sounds as though it hasn't yet replaced `glibc`'s RPC implementation. I am :+1: for this patch, but I don't really have time to test it out until I'm back from traveling in the new year.

DoomHammer:
Is there something I can do to help?

sjackman:
Yeah, can you report on which distro you've tested a Standalone Installation?
Can you try building this Docker image with your change?
https://github.com/sjackman/docker-linuxbrew/blob/master/linuxbrew-gcc-deps/Dockerfile
https://github.com/sjackman/docker-linuxbrew/blob/master/linuxbrew-standalone/Dockerfile

sjackman:
Is packaging TI-RPC for Linuxbrew an alternative solution?

DoomHammer:
I tested on `ubuntu/trusty64` Vagrant image using the [Ubuntu script](https://github.com/Homebrew/linuxbrew/wiki/Standalone-Installation#ubuntu-1404-specific-tldr-setup-script). Tried to build linuxbrew-standalone Docker images but something broke around linuxbrew-gcc-deps (not related) and I surrendered and fell asleep. Will try again if this will help :)

sjackman:
I'm hesitant to modify `glibc` before I leave for traveling for two months. I think it's best to wait until the new year.

DoomHammer:
Understood. BTW: it might also be a good time to update `glibc` version. New year, new glibc, new hopes.

And quite off the main topic: to ensure better quality of Linuxbrew it might be worth considering some automation in this regard. Travis seems like an obvious choice in the beginning, but timeouts would kill us for anything complicated (think `gcc`). I remember reading somewhere that somebody offered hosting space for Jenkins use. Maybe this would help?

sjackman:
A new version of `glibc` would be good too. I've never updated `glibc` so, we'll have to figure out what the upgrade routine is for Standalone users. I seem to recall that the newest version of `glibc` does not work on CentOS 5, because the kernel is too old (2.6.18). I'll have to package old versions of `glibc`, I think.

I'm all for CI testing. (and maybe automated bottling?) I haven't had the time to implement it.

DoomHammer:
You may take a look at #601 where I ported some Travis CI code from both
Homebrew and homebrew-science.

czw., 15.10.2015 o 17:42 użytkownik Shaun Jackman <notifications@github.com>
napisał:

sjackman:
Would packaging TI-RPC for Linuxbrew work for you?

DoomHammer:
May try it when I find a bit of spare time :)

sjackman:
Great. Tell me how it works out for you.

DoomHammer:
Tried to use Docker images but they are month old. Tried to generate my own then, but `gcc-5.3.0` won't compile. Should I open new issue for it?

sjackman:
I'm getting a weird Docker error that's preventing updating the Docker images. See https://github.com/docker/docker/issues/19309
If your error is different then open a new issue.

DoomHammer:
My problem is Linuxbrew-specific and deals with upgrading `gcc`. I believe I have seen it mentioned somewhere around here but not sure whether it have been already fixed?

sjackman:
What's the error message?

DoomHammer:
The error was related to upgrading `gcc` and I believe has been addressed already somewhere else. I have packaged TI-RPC but since I don't have currently access to the mentioned system I couldn't test it.

If TI-RPC helps, we can close this issue.

sjackman:
Does TI-RPC resolve your original use case of compiling `libvirt`? If so, can you submit a pull request?

DoomHammer:
I would have to get access to the system I was using back then. I am not sure when would this happen, but I will test it.

sjackman:
When I try to compile `libtirpc` I get this error message:
```
libtool: compile:  /gsc/btl/linuxbrew/bin/gcc-5 -DHAVE_CONFIG_H -I. -I.. -I../tirpc -include config.h -DPORTMAP -DINET6 -D_GNU_SOURCE -Wall -pipe -isystem/gsc/btl/linuxbrew/include -DHAVE_RPCSEC_GSS -I/gsc/btl/linuxbrew/Cellar/krb5/1.14/include -Os -w -pipe -march=core2 -c rpc_soc.c  -fPIC -DPIC -o .libs/libtirpc_la-rpc_soc.o
In file included from /gsc/btl/linuxbrew/include/nis/rpcsvc/nis.h:520:0,
                 from /gsc/btl/linuxbrew/include/rpcsvc/nis.h:1,
                 from rpc_soc.c:64:
/gsc/btl/linuxbrew/include/rpcsvc/nislib.h: In function 'libnsl_hidden_proto':
/gsc/btl/linuxbrew/include/rpcsvc/nislib.h:5:1: error: expected declaration specifiers before 'libnsl_hidden_proto'
 libnsl_hidden_proto (nis_name_of_r)
 ^
In file included from /gsc/btl/linuxbrew/include/rpcsvc/nis.h:1:0,
                 from rpc_soc.c:64:
/gsc/btl/linuxbrew/include/nis/rpcsvc/nis.h:541:22: error: storage class specified for parameter 'nis_lookup_3'
 extern  nis_result * nis_lookup_3 (ns_request *, CLIENT *) __THROW;
                      ^
```

sjackman:
Here's the formula that I'm using
```ruby
class Libtirpc < Formula
  desc "A port of Suns Transport-Independent RPC library to Linux."
  homepage "http://sourceforge.net/projects/libtirpc/"
  url "http://downloads.sourceforge.net/project/libtirpc/libtirpc/1.0.1/libtirpc-1.0.1.tar.bz2"
  sha256 "5156974f31be7ccbc8ab1de37c4739af6d9d42c87b1d5caf4835dda75fcbb89e"
  # tag "linuxbrew"

  depends_on "homebrew/dupes/krb5" unless OS.mac?

  def install
    system "./configure",
      "--disable-debug",
      "--disable-dependency-tracking",
      "--disable-silent-rules",
      "--prefix=#{prefix}"
    system "make", "install"
  end
end
```

sjackman:
I've created a pull request for `libtirpc` https://github.com/Linuxbrew/linuxbrew/pull/765

sjackman:
`libtirpc` doesn't appear to be a drop-in replacement for `glibc`'s RPC:
```
$ brew list libtirpc
/home/linuxbrew/.linuxbrew/Cellar/libtirpc/1.0.1/etc/netconfig
/home/linuxbrew/.linuxbrew/Cellar/libtirpc/1.0.1/include/tirpc/ (34 files)
/home/linuxbrew/.linuxbrew/Cellar/libtirpc/1.0.1/lib/pkgconfig/libtirpc.pc
/home/linuxbrew/.linuxbrew/Cellar/libtirpc/1.0.1/lib/ (4 files)
/home/linuxbrew/.linuxbrew/Cellar/libtirpc/1.0.1/share/man/ (36 files)
```

sjackman:
I've added `--enable-obsolete-rpc` to `glibc`. Thanks, Piotr.

DoomHammer:
Ha, great, thanks :)

PR #605, opened 2015-10-15
Ensure gpatch when patch step exists.

DoomHammer:
When installing Linuxbrew Standalone on an almost bare system (think embedded-ish here) `patch` may not be available. Add it to dependencies when needed.

### DIFF
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -145,6 +145,7 @@ class SoftwareSpec
   end
 
   def patch(strip = :p1, src = nil, &block)
+    dependency_collector.add("homebrew/dupes/gpatch") unless OS.mac?
     patches << Patch.create(strip, src, &block)
   end
 


sjackman:
Thanks for the patch, Piotr.

sjackman:
Hi, Piotr. Although I agree with this patch in principle, it creates a cyclic dependency. Because `gpatch` is in the tap `homebrew/dupes` it requires `git` to build, and if `git` or any of its dependencies have a patch, it'll require `gpatch` to build.

DoomHammer:
What can be done then? Bottling it?

sjackman:
Yeah, bottling it is a good idea. I can't update the formula though because `gpatch` lives in `Homebrew/dupes`, and I don't have write access to `Homebrew/dupes`. Also bottling may break the dependency cycle for the common case, but the issue will still come up when bootstrapping or on a system that doesn't use bottles. My recommendation is to add
`brew install https://raw.githubusercontent.com/Homebrew/homebrew-dupes/master/gpatch.rb`
to your installation instructions.

DoomHammer:
Seems fair.

sjackman:
I've thought a bit more, and I think this patch is a good plan, but it needs two things. 1. A `PatchRequirement` class to test whether `patch` is already in the `PATH`, in which case `gpatch` is not necessary, and 2. I'll copy `gpatch.rb` from `Homebrew/dupes` into Linuxbrew to prevent the dependency cycle described above.

See [GitRequirement](https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/requirements.rb#L130-L134) for what `PatchRequirement` should look like.

DoomHammer:
One question regarding `PatchRequrement` way: this will mean `gpatch` won't be installed when a system-wide `patch` is available. But isn't the current idea to have Linuxbrew as standalone as possible?

sjackman:
Yeah, I changed my mind about that too. It's in the core repository, and it's bottled, so the `PatchRequirement` isn't really needed.

DoomHammer:
Ah, cool. I tried to implement it but was running into problems, so I like your current state of mind ;)

PR #612, opened 2015-10-21
LinuxBrew with PIE: Permissions get wrongly 'cleaned'

thriqon:
**tl;dr**: `alpine$ brew install hello && brew test hello` fails with insufficient permissions due to PIE.

Alpine Linux (and possibly others) produce [Position Independent Executables](https://en.wikipedia.org/wiki/Position-independent_code), which take the form of shared libraries.
However, from an ELF and therefore from a `path.mach_o_executable?` point of view those are not really `executable`s, which leads to them wrongfully getting `chmod 0444`ed.

## Steps to reproduce

    cat > Dockerfile <<EOF
    FROM alpine
    RUN apk update &&\
      apk add git make clang ruby ruby-irb ncurses tar binutils build-base &&\
      git clone https://github.com/Homebrew/linuxbrew.git ~/brew
    RUN /root/brew/bin/brew install hello
    RUN /root/brew/bin/brew test hello
    EOF
    docker build .

## Suggestion to fix

There is no harm in having libraries executable. In fact, they usually _are_ marked executable anyways. I suggest expanding the check in [Library/Homebrew/cleaner.rb:90](https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/cleaner.rb#L90)

```ruby
perms = if path.mach_o_executable? || path.text_executable?
```

to

```ruby
perms = if path.mach_o_executable? || path.dylib? || path.text_executable?
```

EDIT: I'm unable to run the tests on my machine... It seems to expect to be run on a Mac?

### DIFF
--- a/Library/Homebrew/cleaner.rb
+++ b/Library/Homebrew/cleaner.rb
@@ -87,7 +87,7 @@ class Cleaner
         path.unlink
       else
         # Set permissions for executables and non-executables
-        perms = if path.mach_o_executable? || path.text_executable?
+        perms = if path.mach_o_executable? || path.dylib? || path.text_executable?
           0555
         else
           0444
--- a/Library/Homebrew/test/test_cleaner.rb
+++ b/Library/Homebrew/test/test_cleaner.rb
@@ -23,9 +23,9 @@ class CleanerTests < Homebrew::TestCase
     Cleaner.new(@f).clean
 
     assert_equal 0100555, (@f.bin/"a.out").stat.mode
-    assert_equal 0100444, (@f.lib/"fat.dylib").stat.mode
-    assert_equal 0100444, (@f.lib/"x86_64.dylib").stat.mode
-    assert_equal 0100444, (@f.lib/"i386.dylib").stat.mode
+    assert_equal 0100555, (@f.lib/"fat.dylib").stat.mode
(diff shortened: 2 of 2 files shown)


thriqon:
Shall I request this on Homebrew/homebrew instead?

DoomHammer:
I believe this belongs here and is of not much value to Homebrew. @sjackman is travelling right now so might be pretty slow reviewing PRs.

sjackman:
That's unusual to use PIE for executables. Why does Alpine do that?

I don't see any harm in applying this patch, but it'll have to wait until I'm back from vacation in the new year. Feel free to bump this thread then.

thriqon:
PIE is used for Address Space Layout Randomization. It's used to protect against attacks based on knowledge of the memory layout (buffer overflows etc). AFAIK the position of `main` is shifted around as well, which makes it impossible to use 'normal' executables.

I whish you a nice travel! I'll bump it again then...

sjackman:
Ah, yes. I actually thought that was more common even with Redhat and Debian. Perhaps only for shared libraries.

sjackman:
Merged, making sure to keep the behaviour unchanged on Mac OS: dylibs should be executable on Linux but not on Mac. Thanks for your contribution, Jonas!

PR #622, opened 2015-10-30
llvm: Include the libcxxabi component

dholm:
This is required for `rtags` to build.


sjackman:
Is libcxxabi useful on Mac as well? If so, please submit this patch upstream to Homebrew/homebrew.

sjackman:
Ping? The [libcxxabi homepage](http://libcxxabi.llvm.org/) indicates that it does work on Darwin. Should this patch go upstream to Homebrew/homebrew?

PR #635, opened 2015-11-08
Update centos installation instructions

PerilousApricot:
For centos 6.7 (at least), the "ruby" package doesn't pull in irb. Update instructions for centos to pull it in.


PerilousApricot:
FWIW, I'm using homebrew to install some newer versions of packages than what centos provides. Centos6' EOL is in 2020 (!), so having somewhat new development tools is a necessity.

sjackman:
index.html is auto generated. Please modify README.md instead.

sjackman:
Fixed,

PR #636, opened 2015-11-08
Make sure irb is installed

PerilousApricot:
Under centos 6.7 (at least), installing the "ruby" package doesn't install the irb module. Warn the user in this case.


sjackman:
Thanks for this patch. Could you instead add it to the dependencies section of README.md?

sjackman:
`yum install irb` is in the RedHat installation instructions:
https://github.com/Homebrew/linuxbrew/blob/master/README.md#fedora-centos-or-red-hat

PR #643, opened 2015-11-13
boost.rb with gcc now compiles boost.log on linux

ahundt:
eventually it would be ideal to only enable this if the gcc version is >=4.5
https://github.com/Homebrew/homebrew/issues/45954

### DIFF
--- a/Library/Formula/boost.rb
+++ b/Library/Formula/boost.rb
@@ -88,7 +88,7 @@ class Boost < Formula
 
     # Boost.Log cannot be built using Apple GCC at the moment. Disabled
     # on such systems.
-    without_libraries << "log" if ENV.compiler == :gcc || ENV.compiler == :llvm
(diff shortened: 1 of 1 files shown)


sjackman:
Thanks for the bug report, Andrew. `brew install gcc boost` should also work as a workaround.

ahundt:
brew install gcc will install the homebrew gcc, and I'd prefer to use the system gcc if possible, correct?

sjackman:
Yes, correct. I'm on vacation until the new year. Please bump this thread after then. Sorry for the delay.

sjackman:
Which Linux distribution are you using, and which version of GCC does it include?
Creating versioned GCC symlinks should resolve your issue.
https://github.com/Homebrew/linuxbrew/wiki/Standalone-Installation#create-symlinks-for-gcc

sjackman:
`ENV.compiler == :gcc` in Homebrew implies GCC 4.2. If you use a versioned symlink as your Homebrew compiler, such as `g++-4.8`, it should resolve this issue.

PR #645, opened 2015-11-13
apache-spark 1.5.2

youtux:


youtux:
Sorry, I assumed that the contributing guidelines were the same as for Homebrew. I was wrong.

PR #653, opened 2015-11-17
Fixed OS detection on Linux reporting "java" instead of "linux"

wesleybliss:
Some Linux distros come with JRuby pre-installed (e.g. Fedora). Since I never use Ruby, I never noticed, but when the installer failed, reporting "java" as the "unknown os", I realized the RUBY_PLATFORM constant always reports "java" for JRuby - regardless of the actual host OS.

This just adds an extra check for Linux. You may consider not using RUBY_PLATFORM at all, but that's up to you.

### DIFF
--- a/Library/Homebrew/os.rb
+++ b/Library/Homebrew/os.rb
@@ -1,10 +1,12 @@
+require 'rbconfig'
+
 module OS
   def self.mac?
     /darwin/i === RUBY_PLATFORM
   end
 
   def self.linux?
-    /linux/i === RUBY_PLATFORM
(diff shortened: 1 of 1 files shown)


sjackman:
Interesting. Homebrew doesn't support I believe any other ruby than the standard ruby. Does JRuby work just fine otherwise after this patch? What's the motivation for using JRuby instead of Ruby?
I'm on vacation in Thailand until the new year. I won't be able to address this issue until I return. Sorry for the delay.

sjackman:
Merged. Thanks for your contribution, Wesley.

PR #671, opened 2015-11-30
Fix erroneous MySQL dependencies

peteygao:
Remove duplicate `depends_on` declarations for `cmake`, `pidof` and `openssl`. Specifically the duplicate `depends_on pidof` without the `!OS.mac?` guard was breaking the build on non-Mac OS's. See line 15 of the same file for the correct declaration.

This fixes the issue I filed: https://github.com/Homebrew/linuxbrew/issues/670

However it seems like the build _still_ will not complete due to some issue with Boost (which will now attempt to build without the `pidof` failure). I'll file a separate issue for that.

### DIFF
--- a/Library/Formula/mysql.rb
+++ b/Library/Formula/mysql.rb
@@ -16,6 +16,7 @@ class Mysql < Formula
   depends_on "openssl"
   # Fix error: Cannot find system editline libraries.
   depends_on "libedit" unless OS.mac?
+  depends_on "boost"
 
   option :universal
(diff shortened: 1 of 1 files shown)


DomT4:
Apologies, I missed the fact you'd filed a PR for this and committed the fix in https://github.com/Homebrew/linuxbrew/commit/b82c4cf21c9a9e70511babfc8c03dad932fce6c1.

Feel free to open an issue for the Boost problem. Thanks for doing some debugging on the errors here, appreciate that!

PR #672, opened 2015-11-30
Update Boost to 1.59.0

peteygao:
I noticed that this was reverted in a revert commit two weeks ago. It was part of a huge changeset, so I'm guessing it was unintentional for Boost in particular to be reverted (but if it _was_ intentional, feel free to say so and close this PR).

This is needed as the latest MySQL formula requires Boost 1.59.0.

### DIFF
--- a/Library/Formula/boost.rb
+++ b/Library/Formula/boost.rb
@@ -1,17 +1,17 @@
 class Boost < Formula
   desc "Collection of portable C++ source libraries"
   homepage "http://www.boost.org"
-  url "https://downloads.sourceforge.net/project/boost/boost/1.58.0/boost_1_58_0.tar.bz2"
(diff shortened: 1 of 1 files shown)


DomT4:
Welp. Looks like the stuff we had to revert a couple weeks ago didn't then get dragged in with the new merge @danabrand. Any thoughts on that? Maybe we could try reverting the reverted merges, if we've addressed the `os/mac` issue we encountered there.

danabrand:
@DomT4 Damn I thought the mess was cleaned up. Let me look at it.

ahundt:
I'm also having problems because boost.log isn't built, which is supported on these OS versions and something I depend on.

peteygao:
@DomT4 or @danabrand Bump.

DomT4:
Shaun is back now and working on sorting Linuxbrew out. I am thankfully back to "only" maintaining Homebrew, heh.

sjackman:
I've merged Homebrew into Linuxbrew, which updated boost to 1.59.0

PR #675, opened 2015-12-03
libpng 1.6.19: Fix for linuxbrew, link the package, otherwise packages

paulhybryant:
libpng 1.6.19: Fix for linuxbrew, link the package, otherwise packages
depending on libpng will failed to build.

### DIFF
--- a/Library/Formula/libpng.rb
+++ b/Library/Formula/libpng.rb
@@ -12,7 +12,7 @@ class Libpng < Formula
     sha256 "39ba75af10d801cec077b3fc064b5011663cf4ad16cbbfe9b7f5c33fddef27e4" => :mavericks
   end
 
-  keg_only :provided_pre_mountain_lion
+  keg_only :provided_pre_mountain_lion unless OS.linux?
 
   head do
     url "https://github.com/glennrp/libpng.git"


DomT4:
I thought `keg_only` was ignored for Linuxbrew anyway? @sjackman

paulhybryant:
It seems it is not ignored. When I try to upgrade / install a formula that depends on libpng, I run into this issue.

DomT4:
What do you see, specifically? Linux shouldn't even be reading `provided_pre_mountain_lion` because Mountain Lion is definitely not a Linux distro release :sweat_smile:.

paulhybryant:
ok, this is what I just did.

$ brew uninstall libpng
$ brew install libpng (this is with the unchanged libpng.rb)

==> Downloading
ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/libpng-1.6.19.tar.xz
Already downloaded: /home/yuhuang/.cache/Homebrew/libpng-1.6.19.tar.xz
==> ./configure --disable-silent-rules
--prefix=/home/yuhuang/.linuxbrew/Cellar/libpng/1.6.19
==> make
==> make test
==> make install
==> Caveats
This formula is keg-only, which means it was not symlinked into
/home/yuhuang/.linuxbrew.

OS X already provides this software in versions before Mountain Lion.

Generally there are no consequences of this for you. If you build your
own software and it requires this formula, you'll need to add to your
build variables:

    LDFLAGS:  -L/home/yuhuang/.linuxbrew/opt/libpng/lib
    CPPFLAGS: -I/home/yuhuang/.linuxbrew/opt/libpng/include


==> Summary
/home/yuhuang/.linuxbrew/Cellar/libpng/1.6.19: 17 files, 1.3M, built in 27
seconds

$ vi
vi: error while loading shared libraries: libpng16.so.16: cannot open
shared object file: No such file or directory

$ brew link --force libpng
Linking /home/yuhuang/.linuxbrew/Cellar/libpng/1.6.19... 19 symlinks

$ vi
(now everything works fine)

On Fri, Dec 4, 2015 at 1:40 PM, Dominyk Tiller <notifications@github.com>
wrote:

> What do you see, specifically? Linux shouldn't even be reading
> provided_pre_mountain_lion because Mountain Lion is definitely not a
> Linux distro release [image: :sweat_smile:].
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Homebrew/linuxbrew/pull/675#issuecomment-161881747>.
>



-- 
With Regards,

Yu Huang

Software Engineer,
Google Inc.

sjackman:
Looks like a bug in Linuxbrew to me. I'm on vacation and back in the new year.

sjackman:
I can't reproduce this issue. Possibly it was fixed in the last merge.

sjackman:
This case is handled in the core code here: https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/formula_support.rb#L16-L17

PR #685, opened 2015-12-18
moz-git-tools 0.1 (import from head-only)

rwhogg:
This imports moz-git-tools version 0.1 into Homebrew core.

I haven't made the commit to remove it from head-only yet; I will soon.


rwhogg:
Oh shoot, this was supposed to go to Homebrew, not Linuxbrew. My bad!

PR #694, opened 2015-12-29
Libevent 404

bloomyminded:
Libevent url was giving me a 404, updated with the latest release from sourceforge.


sjackman:
This issue will be resolved when Homebrew is merged into Linuxbrew in the new year.

PR #698, opened 2016-01-02
uniformed syntax

retokromer:
always the same "if-then" syntax


sjackman:
Please send this patch upstream to https://github.com/Homebrew/homebrew

PR #717, opened 2016-01-14
android-ndk: Use Linux binaries on linxbrew

chrissicool:
The original Homebrew version of this formula uses the prebuilt Darwin
binaries. This patch downloads and unpacks the Linux binaries instead
when we are not on a Mac.


sjackman:
Thanks for your contribution, Chris.

PR #726, opened 2016-01-18
rtags: fix on linux - adjust for libc++abi only when on OS/X

jeffkowalski:
The adjustment ```ENV.append("LDFLAGS", "-lc++abi")``` is necessary only on OSX, _not_ linux, where it will generate an error.
The fix is to execute that statement conditionally,
```
    # Homebrew llvm libc++.dylib doesn't correctly reexport libc++abi
    ENV.append("LDFLAGS", "-lc++abi") if OS.mac?
```

### DIFF
--- a/Library/Formula/rtags.rb
+++ b/Library/Formula/rtags.rb
@@ -19,7 +19,7 @@ class Rtags < Formula
 
   def install
     # Homebrew llvm libc++.dylib doesn't correctly reexport libc++abi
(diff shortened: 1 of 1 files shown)


sjackman:
Merged. Thanks for your contribution, Jeff!

For future pull requests, the preferred commit log message is
`rtags: Fix for Linuxbrew`

PR #728, opened 2016-01-19
cairo: add X11 support

Rudloff:
Hello,

The cairo formula currently can't be built on Linux with the `--with-x11` option.
This patch adds xlib support.

### DIFF
--- a/Library/Formula/cairo.rb
+++ b/Library/Formula/cairo.rb
@@ -16,7 +16,7 @@ class Cairo < Formula
   option :universal
 
   depends_on "pkg-config" => :build
-  depends_on :x11 => :optional if MacOS.version > :leopard
+  depends_on :x11 => :optional if MacOS.version > :leopard || OS.linux?
   depends_on "freetype"
   depends_on "fontconfig"
   depends_on "libpng"
@@ -38,7 +38,11 @@ class Cairo < Formula
     ] if OS.mac?
 
     if build.with? "x11"
-      args << "--enable-xcb=yes" << "--enable-xlib=yes" << "--enable-xlib-xrender=yes"
+      if OS.linux?
+        args << "--enable-xcb=no" << "--enable-xlib=yes" << "--enable-xlib-xrender=yes"
+      else
+        args << "--enable-xcb=yes" << "--enable-xlib=yes" << "--enable-xlib-xrender=yes"
+      end
     else
       args << "--enable-xcb=no" << "--enable-xlib=no" << "--enable-xlib-xrender=no"
     end


paulhybryant:
Curious, I have done this before as well. Do you have to fake the x11 packages? Because the pkgconfig files for e.g. xcb cannot be found otherwise.

sjackman:
What error message are you seeing? Please report `brew gist-logs cairo`
Does it help to `sudo apt-get install libx11-xcb-dev`?

Rudloff:
I don't get any error message. Without my patch, cairo is simply built without xlib support (the option is not available).
Then I get this kind of error when I try to build formulae like gtk+:
```No package 'cairo-xlib' found```

**Comment by sjackman** [comment]
Ah, I see. Does installing `libx11-xcb-dev` help?

**Comment by Rudloff** [comment]
It is already installed.

**Comment by sjackman** [comment]
I'd prefer to figure out why `xcb

**Comment by Rudloff** [comment]
Here it is: https://gist.github.com/Rudloff/ebd6baa383b0e64de13f

**Comment by sjackman** [comment]
Sorry, I meant to say include the first chunk `depends_on :x11 => :optional if MacOS.version > :leopard || OS.linux?` but not the second chunk `--enable-xcb=no` and report `brew install cairo --with-x11; brew gist-logs cairo`

**Comment by Rudloff** [comment]
Oh, OK.
In this case, I do get an error:
```
checking for cairo's XCB surface backend feature... 
checking for xcb... no
checking whether cairo's XCB surface backend feature could be enabled... no (requires xcb >= 1.6 xcb-render >= 1.6 http://xcb.freedesktop.org)
configure: error:  XCB surface backend feature could not be enabled
```

Here are the logs: https://gist.github.com/Rudloff/229344610460955b98fd

**Comment by sjackman** [comment]
Does `sudo apt-get install libxcb-render0-dev` help?
What version of `xcb` and `xcb-render` do you have installed?

**Comment by Rudloff** [comment]
It is already installed (1.10-3+b1).

**Comment by sjackman** [comment]
https://gist.github.com/Rudloff/229344610460955b98fd#file-config-log-L2356-L2357
```
Package xcb was not found in the pkg-config search path.
Perhaps you should add the directory containing `xcb.pc'
```
Set `PKG_CONFIG_PATH` to the directory containing `xcb.pc`.
See also #47

**Comment by paulhybryant** [comment]
Yeah, that is why I ask my question in my reply. I don't think linuxbrew supports the ":x11" dependencies or am I wrong? In the past what I have to do is to change PKG_CONFIG_PATH / PKG_CONFIG_LIB in the formula, but it seems in this PR adding the dependency of ":x11" will work just fine.

**Comment by sjackman** [comment]
Correct. The `:x11` dependency has no effect on Linuxbrew.

**Comment by paulhybryant** [comment]
Thanks for the clarification @sjackman . 

@Rudloff In that case I am wondering how this PR works as the pkgconfig for xcb will not be found from linuxbrew, unless it is manually symlinked somewhere in linuxbrew's pkgconfig search path (is that what you did?), or additional paths are added in PKG_CONFIG_PATH in the formula (which is not the case in this PR)

**Comment by sjackman** [comment]
I've merged the first chunk of this patch.
```difff
-  depends_on :x11 => :optional if MacOS.version > :leopard
+  depends_on :x11 => :optional if MacOS.version > :leopard || !OS.mac?
```
Thank you for your contribution, Pierre.

PR #749, opened 2016-01-26
bin/brew: fix stat on linux

ntkme:
This PR fixes the following error:

```
$ sudo brew
stat: invalid option -- '%'
Try 'stat --help' for more information.
```

`stat` in GNU coreutils uses `-c` for format, where BSD `stat` uses `-f`.


sjackman:
Linuxbrew aims to be portable to both Linux and Mac OS, and `stat -c %u` does not work on Mac OS.
I went with `ls -ld |awk '{print $3}'`. I'm not keen that it requires awk.
See http://unix.stackexchange.com/questions/7730/find-the-owner-of-a-directory-or-file-but-only-return-that-and-nothing-else

PR #755, opened 2016-01-26
Fix libffi: make universal binary for OS X only

maxim-belkin:
Without this fix, building libffi from source fails on Linux with an error message: "Error: Non-Apple GCC can't build universal binaries"

### DIFF
--- a/Library/Formula/libffi.rb
+++ b/Library/Formula/libffi.rb
@@ -34,8 +34,8 @@ class Libffi < Formula
     # Move lib64/* to lib/ on Linuxbrew
     lib64 = Pathname.new "#{lib}64"
     if lib64.directory?
-      system "mv #{lib64}/* #{lib}/"
-      rmdir lib64
+      mv Dir.glob('#{lib64}/*'), '#{lib}'
+      rmdir '#{lib64}'
     end
   end
 
--- a/Library/Homebrew/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/ENV/std.rb
@@ -294,6 +294,7 @@ module Stdenv
   end
 
   def universal_binary
+    return unless OS.mac?
     check_for_compiler_universal_support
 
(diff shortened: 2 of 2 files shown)


sjackman:
This issue should be resolved by disabling the `ENV.universal_binary` function:
https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/extend/ENV/std.rb#L296
Add `return unless OS.mac?`

maxim-belkin:
`return unless OS.mac?` stops compilation on Linux at all.

sjackman:
Sorry, I wasn't clear. Add `return unless OS.mac?` to the function `ENV.universal_binary` in `std.rb`.

maxim-belkin:
Thanks! I've added the proposed change to std.rb and updated my request.

Library/Formula/libffi.rb
```diff
@@ -34,8 +34,8 @@ def install
     # Move lib64/* to lib/ on Linuxbrew
     lib64 = Pathname.new "#{lib}64"
     if lib64.directory?
-      system "mv #{lib64}/* #{lib}/"
-      rmdir lib64
+      mv Dir.glob('#{lib64}/*'), '#{lib}'
+      rmdir '#{lib64}'
```
sjackman: Please revert this chunk. Is it purely an aesthetic change?

sjackman:
Please squash all the commits to a single commit and change the commit log message to "Disable ENV.universal_binary for Linuxbrew".

Library/Formula/libffi.rb
maxim-belkin: Without this change `brew audit libffi` complains about a call to system's `mv` command

maxim-belkin:
Updated my request as requested :)

sjackman:
Thanks for your contribution, Maxim.

PR #756, opened 2016-01-27
nspr: Fix for Linuxbrew

rwhogg:
Note that this *might* be susceptible to the issue described at https://github.com/Homebrew/homebrew/issues/3041 . It's impossible to tell, since neither spidermonkey nor couchdb install at the moment.

### DIFF
--- a/Library/Formula/nspr.rb
+++ b/Library/Formula/nspr.rb
@@ -37,7 +37,7 @@ class Nspr < Formula
       args << "--enable-64bit" if MacOS.prefer_64_bit?
       system "./configure", *args
       # Remove the broken (for anyone but Firefox) install_name
-      inreplace "config/autoconf.mk", "-install_name @executable_path/$@ ", "-install_name #{lib}/$@ "
+      inreplace "config/autoconf.mk", "-install_name @executable_path/$@ ", "-install_name #{lib}/$@ " if OS.mac?
 
       system "make"
       system "make", "install"


Library/Formula/nspr.rb
```diff
@@ -32,12 +34,12 @@ def install
         --enable-strip
         --enable-pthreads
         --enable-ipv6
-        --enable-macos-target=#{MacOS.version}
       ]
-      args << "--enable-64bit" if MacOS.prefer_64_bit?
+      args << "--enable-macos-target=#{MacOS.version}" if OS.mac?
+      args << "--enable-64bit" if MacOS.prefer_64_bit? || OS.linux?
```
rwhogg: I'm guessing based on the changes that were made to the sonar formula that you probably don't want the `OS.linux?` bit here.

I'll happily modify it tomorrow if you'd like, but I'm tired so I'm going to leave it for now.

Library/Formula/nspr.rb
```diff
@@ -32,12 +34,12 @@ def install
...
       system "./configure", *args
       # Remove the broken (for anyone but Firefox) install_name
-      inreplace "config/autoconf.mk", "-install_name @executable_path/$@ ", "-install_name #{lib}/$@ "
+      inreplace "config/autoconf.mk", "-install_name @executable_path/$@ ", "-install_name #{lib}/$@ " if OS.mac?
```
rwhogg: This is the line that might trigger the couchdb/spidermonkey bug I mentioned in the PR description. I've just disabled it entirely now, since I couldn't find any instance of "-install_name" in `autoconf.mk` and the line fails otherwise.

We might have to fix this later when we fix the spidermonkey build, but IMO it's fine for now.

Thoughts?

rwhogg:
The Travis build failed for the same reason Sonar did.

Library/Formula/nspr.rb
```diff
@@ -21,10 +21,12 @@ class Nspr < Formula
...
-      # Fixes a bug with linking against CoreFoundation, needed to work with SpiderMonkey
-      # See: http://openradar.appspot.com/7209349
-      target_frameworks = (Hardware.is_32_bit? || MacOS.version <= :leopard) ? "-framework Carbon" : ""
-      inreplace "pr/src/Makefile.in", "-framework CoreServices -framework CoreFoundation", target_frameworks
+      if OS.mac?
+        # Fixes a bug with linking against CoreFoundation, needed to work with SpiderMonkey
+        # See: http://openradar.appspot.com/7209349
+        target_frameworks = (Hardware.is_32_bit? || MacOS.version <= :leopard) ? "-framework Carbon" : ""
+        inreplace "pr/src/Makefile.in", "-framework CoreServices -framework CoreFoundation", target_frameworks
+      end
```
sjackman: Is this chunk necessary? It looks as though it shouldn't have any effect on Linux.

Library/Formula/nspr.rb
sjackman: Correct. `args << "--enable-64bit" if MacOS.prefer_64_bit?` should be fine.

Library/Formula/nspr.rb
sjackman: `-install_name` has no effect on Linux. This chunk should not be necessary. The equivalent option on Linux is `-Wl,-soname`

Library/Formula/nspr.rb
rwhogg: It probably isn't. I can try taking it out next time.

sjackman:
I think that the only chunk that should be necessary is `args << "--enable-macos-target=#{MacOS.version}" if OS.mac?`. Does that line alone suffice?

Library/Formula/nspr.rb
rwhogg: Sorry it took so long to respond, I've been sick the last few days. Will try to get to this later today.

Library/Formula/nspr.rb
rwhogg: This chunk sadly is necessary, otherwise the `inreplace` fails and stops the build.

Library/Formula/nspr.rb
```diff
@@ -32,12 +32,12 @@ def install
...
         --enable-pthreads
         --enable-ipv6
-        --enable-macos-target=#{MacOS.version}
       ]
+      args << "--enable-macos-target=#{MacOS.version}" if OS.mac?
       args << "--enable-64bit" if MacOS.prefer_64_bit?
       system "./configure", *args
       # Remove the broken (for anyone but Firefox) install_name
-      inreplace "config/autoconf.mk", "-install_name @executable_path/$@ ", "-install_name #{lib}/$@ "
+      inreplace "config/autoconf.mk", "-install_name @executable_path/$@ ", "-install_name #{lib}/$@ " if OS.mac?
```
sjackman: Is this chunk necessary?

rwhogg: Yes. The build fails without it.

I didn't copy down the precise error message, but if you'd like I can uninstall and reinstall without it to show you what it is.

sjackman:
```
Utils::InreplaceError: inreplace failed
config/autoconf.mk:
  expected replacement of "-install_name @executable_path/$@ " with "-install_name /gsc/btl/linuxbrew/Cellar/nspr/4.11/lib/$@ "
```

Library/Formula/nspr.rb
```diff
@@ -32,12 +32,12 @@ def install
         --enable-strip
         --enable-pthreads
         --enable-ipv6
-        --enable-macos-target=#{MacOS.version}
       ]
+      args << "--enable-macos-target=#{MacOS.version}" if OS.mac?
```
sjackman: This chunk appears to be unnecessary. `--enable-macos-target=0` seems to have no ill effect on Linux.

rwhogg: Gotcha, will change.

sjackman:
Merged. Thanks for your contributions, Bob.

PR #760, opened 2016-01-28
sha2: Remove www from upstream hostname

thriqon:
SSL certificate is not valid for `www.aarongifford.com`, only for `aarongifford.com`.


sjackman:
Thanks for your contribution, Jonas. Patches that are not specific to Linux should be sent upstream to [Homebrew](https://github.com/Homebrew/homebrew]. See [CONTRIBUTING.md](https://github.com/Linuxbrew/linuxbrew/blob/master/CONTRIBUTING.md)

DomT4:
Was fixed upstream earlier today FWIW :)

sjackman:
Thanks, Dominyk.
@thriqon Homebrew is merged into Linuxbrew on a weekly basis. I also occasionally do merges on special request. Merged and fixed.

PR #765, opened 2016-01-30
libtirpc 1.0.1 (new formula)

sjackman:
A port of Suns Transport-Independent RPC library to Linux.


sjackman:
See #603

sjackman:
```
==> audit problems
libtirpc:
 * Incorrect file permissions (664): chmod 644 /home/travis/build/Linuxbrew/linuxbrew/Library/Formula/libtirpc.rb
 * A `test do` test block should be added
```

PR #776, opened 2016-02-01
Use util-linux rather than libuuid for Linuxbrew

sjackman:
util-linux provides libuuid and more.

### DIFF
--- a/Library/Formula/libuuid.rb
+++ b/Library/Formula/libuuid.rb
@@ -5,6 +5,7 @@ class Libuuid < Formula
   url "https://downloads.sourceforge.net/project/libuuid/libuuid-1.0.3.tar.gz"
   sha256 "46af3275291091009ad7f1b899de3d0cea0252737550e7919d17237997db5644"
 
+  conflicts_with "libuuid", :because => "both install lib/libuuid.a"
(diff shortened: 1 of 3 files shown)


sjackman:
https://travis-ci.org/Linuxbrew/linuxbrew/builds/106355108#L528-L530
```
brew install --verbose --build-bottle task
Error: No available formula with the name "homebrew/dupes/ncurses" 
Please tap it and then try again: brew tap homebrew/dupes
```

Library/Formula/libuuid.rb
```diff
@@ -5,6 +5,7 @@ class Libuuid < Formula
   url "https://downloads.sourceforge.net/project/libuuid/libuuid-1.0.3.tar.gz"
   sha256 "46af3275291091009ad7f1b899de3d0cea0252737550e7919d17237997db5644"
 
+  conflicts_with "libuuid", :because => "both install lib/libuuid.a"
```
rwhogg: Hey @sjackman - please pardon my commenting on an already-merged pull request, but was this conflicts_with supposed to be in the util-linux formula? The way I see it, the libuuid formula is currently conflicting with itself.

sjackman: Yep. Sure was. Thanks for catching my mistake, Bob.

PR #785, opened 2016-02-04
hello: Test Travis artifacts

sjackman:


sjackman:
To test multiple distros and build architectures:
```yaml
matrix:
  include:
    - sudo: false # dist: precise
    - sudo: required # dist: precise
    - dist: trusty
      sudo: required
```

sjackman:
`ruby 1.9.3p551` fails:
https://travis-ci.org/Linuxbrew/linuxbrew/jobs/115956255#L231-L235
```
Installing simplecov-html 0.10.0
NoMethodError: undefined method `spec' for nil:NilClass
An error occurred while installing simplecov (0.11.2), and Bundler cannot
continue.
```

sjackman:
See https://travis-ci.org/Linuxbrew/linuxbrew/builds/115956252

PR #793, opened 2016-02-06
bzr: Fix Python interpreter for Linuxbrew

bukzor:


Library/Formula/bazaar.rb
```diff
@@ -20,11 +20,8 @@ def install
     system "make", "man1/bzr.1"
     man1.install "man1/bzr.1"
 
-    # Put system Python first in path
-    ENV.prepend_path "PATH", "/System/Library/Frameworks/Python.framework/Versions/Current/bin"
-
     system "make"
-    inreplace "bzr", "#! /usr/bin/env python", "#!/usr/bin/python"
+    inreplace "bzr", "#! /usr/bin/env python", "#!#{which "python"}"
```
sjackman: Use instead
```ruby
inreplace "bzr", "#! /usr/bin/env python", "#!/usr/bin/python" if OS.mac?
```

sjackman:
The preferred commit log message is `bzr: Fix for Linuxbrew`
Please squash to a single commit after this fixes are made (that is, use `git commit --amend`).

Library/Formula/bazaar.rb
bukzor: We don't want `/usr/bin/env python` in the case of linux either, for the same reason that it's not wanted in machomebrew; it's too often that a entirely different python is on the path than the one we compiled with. We should pin it down to the one that's on the path during build.

bukzor:
I believe the travis failure wasn't due to my patch.

bukzor:
This has failed four times in a row. Is master branch broken?
I can't see how my patch would have caused that error.

Library/Formula/bazaar.rb
sjackman: My preference is to trust the user's `PATH`. As a user if I upgrade my Python distribution for example with `brew install python`, I'd prefer that installed packages started using it.

sjackman:
Yes, `master` appears broken. The gem `tins-1.6.0` appears to have gone AWOL.
```
Could not find tins-1.6.0 in any of the sources (Bundler::GemNotFound)
	from /home/travis/.rvm/rubies/ruby-1.9.3-p551/lib/ruby/gems/1.9.1/gems/bundler-1.7.6/lib/bundler/spec_set.rb:85:in `map!'
```

Library/Formula/bazaar.rb
```diff
@@ -20,11 +20,8 @@ def install
     system "make", "man1/bzr.1"
     man1.install "man1/bzr.1"
 
-    # Put system Python first in path
-    ENV.prepend_path "PATH", "/System/Library/Frameworks/Python.framework/Versions/Current/bin"
```
sjackman: Use instead
```ruby
ENV.prepend_path "PATH", "/System/Library/Frameworks/Python.framework/Versions/Current/bin" if OS.mac?
```

sjackman: Or rather leave this line untouched as it causes no harm on Linux.

sjackman:
I've merged
```ruby
inreplace "bzr", "#! /usr/bin/env python", "#!/usr/bin/python" if OS.mac?
```
Thanks for your contribution, Buck.

bukzor:
@sjackman This still results in breakage, since the python on the $PATH during compilation is (sometimes, and in my case) `$BREW/bin/python`.  `#{which}` would give the correct value.

sjackman:
If the `python` formula is installed, wouldn't that be the desired Python interpreter to use, that is `$BREW/bin/python`? Which Python interpreter do you want to use?

bukzor:
Ah, I read it wrong, sorry. You've set it up to be find python using the $PATH under linux. While that fixes my issue (I believe) it's less reliable to pinning to `#{which python}` during the build, since depending on the exact value of $PATH an incompatible python will be used. This is why homebrew pins the shebang; it just happens to be that /usr/bin/python is invariantly the python they compile with.

sjackman:
I understand your and Homebrew's stance, but I prefer not to pin the shebang and to trust the user's `PATH`.

PR #794, opened 2016-02-06
glibc: Add "--enable-kernel" option

maxim-belkin:
glibc: added the "enable-kernel" option which in notes on "Standalone installation" is mentioned as "enable-version".
gmp: added "Include" so that "brew test gmp" succeeds
config.rb: make (which "something") uniform across the entire file + cosmetic stuff

### DIFF
--- a/Library/Formula/glibc.rb
+++ b/Library/Formula/glibc.rb
@@ -4,6 +4,8 @@ class Glibc < Formula
   url "http://ftpmirror.gnu.org/glibc/glibc-2.19.tar.bz2"
   sha256 "2e293f714187044633264cd9ce0183c70c3aa960a2f77812a6390a3822694d15"
 
+  option "with-current-kernel", "Compile for compatibility with kernel not older than your current one"
+
   # binutils 2.20 or later is required
   depends_on "binutils" => [:build, :recommended]
 
@@ -19,6 +21,8 @@ class Glibc < Formula
         "--prefix=#{prefix}",
         "--enable-obsolete-rpc",
         "--without-selinux"] # Fix error: selinux/selinux.h: No such file or directory
+      args << "--enable-kernel=" +
+        `uname -r | cut -d. -f1,2,3 | cut -d- -f1` if build.with? "current-kernel"
       args << "--with-binutils=" +
         Formula["binutils"].bin if build.with? "binutils"
       args << "--with-headers=" +


Library/Formula/glibc.rb
```diff
@@ -4,6 +4,8 @@ class Glibc < Formula
   url "http://ftpmirror.gnu.org/glibc/glibc-2.19.tar.bz2"
   sha256 "2e293f714187044633264cd9ce0183c70c3aa960a2f77812a6390a3822694d15"
 
+  option "enable-kernel", "Compile for compatibility with kernel not older than your current one" if OS.linux?
```
sjackman: `glibc` is a Linux-only package, so the `if OS.linux?` guards are not necessary in this case.

Library/Formula/glibc.rb
```diff
@@ -19,6 +21,9 @@ def install
         "--prefix=#{prefix}",
         "--enable-obsolete-rpc",
         "--without-selinux"] # Fix error: selinux/selinux.h: No such file or directory
+      unless OS.mac?
```
sjackman: Ditto.

Library/Formula/gmp.rb
```diff
@@ -45,7 +45,7 @@ def install
         return 0;
       }
     EOS
-    system ENV.cc, "test.c", "-L#{lib}", "-lgmp", "-o", "test"
+    system ENV.cc, "test.c", "-L#{lib}", "-I#{include}", "-lgmp", "-o", "test"
```
sjackman: Stylistically, I'd prefer
```ruby
system ENV.cc, "-I#{include}", "test.c", "-L#{lib}", "-lgmp", "-o", "test"
```
As this patch is not specific to Linux, please send it upstream to https://github.com/Homebrew/homebrew
See https://github.com/Linuxbrew/linuxbrew/blob/master/CONTRIBUTING.md

Library/Homebrew/cmd/config.rb
```diff
@@ -131,8 +131,9 @@ def describe_system_ruby
 
   def describe_java
     unless OS.mac?
-      return "N/A" if which("java").nil?
-      java = `java -version 2>&1`.lines.first.chomp
+      return "N/A" if (which "java").nil?
+      java = `java -version 2>&1`
+      java = java.lines.first.chomp
```
sjackman: What is the purpose of this chunk?

Library/Formula/glibc.rb
maxim-belkin: I didn't bother checking if glibc exists in homebrew so added this without any deep thought.

Library/Homebrew/cmd/config.rb
maxim-belkin: I guess, stylistic. Before my "standalone" build was finished, line ```java -version 2>&1`.lines.first.chomp`` was failing with an error message (which I don't remember exactly) about `chomp` method being applied to nil:Nilclass object. While troubleshooting the problem, I saw that `(which "<something>")` is more commonly used... so this is the history behind the first change. The rest of the changes were just my attempts to troubleshoot the problem and understand why the same line was working in `irb` but failing when I was running `brew install...`

Library/Formula/glibc.rb
maxim-belkin: Same thing. So, what is the best course of action now? Should I close this pull request and submit a new one? Or were you going to fix it and merge it without my interference?

sjackman: Please fix it up. Make your changes, then commit them using `git commit --amend Library/Formula/glibc.rb` and push them using `git push --force`. Your pull request will be updated.

Library/Homebrew/cmd/config.rb
sjackman: If it's just stylistic, please remove this chunk. Did you figure out what the root cause of the error was?

maxim-belkin: Not really... when the standalone build completed, this error was gone. I've recorded all the steps I was taking so can repeat them if someone else comes across this error and solution will be needed. 
Anyway, I will remove this hunk and the one for `gmp` + fix the glibc formula and update the request.

maxim-belkin:
I've updated the request

sjackman:
This patch won't work as is. `build.with?` assumes the option is named either `--with-foo` or `--without-foo`. The option will have to be renamed to something like `--with-this-kernel`.

maxim-belkin:
fixed!

Library/Formula/glibc.rb
```diff
@@ -19,6 +21,8 @@ def install
         "--prefix=#{prefix}",
         "--enable-obsolete-rpc",
         "--without-selinux"] # Fix error: selinux/selinux.h: No such file or directory
+      args << "--enable-kernel=" +
+        `uname -r | cut -d. -f1,2,3 | cut -d- -f1` if build.with? "current-kernel"
```
sjackman: How about
```ruby
args << "--enable-kernel=#{`uname -r | sed 's/-.*//'`}" if build.with? "current-kernel"
```

maxim-belkin: That's one pipe instead of two - that's a plus. It relies on a dash in the kernel version -- that's a minus. From [here](https://www.kernel.org/category/releases.html)

> It is easy to tell if you are running a distribution kernel. Unless you downloaded, compiled and installed your own version of kernel from kernel.org, you are running a distribution kernel. To find out the version of your kernel, run uname -r:
> 
> $ uname -r
> 3.7.5-201.fc18.x86_64
> 
> **If you see anything at all after the dash**, you are running a distribution kernel. Please use the support channels offered by your distribution vendor to obtain kernel support.
> 

So, relying on a dash in the return value of `uname -r` we put aside those two and a half enthusiasts who compile Linux kernel themselves. But I'm fine with this version if you insist because if someone compiles a kernel - he or she can figure out how to fix this formula as well.

sjackman: What does `uname -r` look like on a non-distribution kernel? If it's simply `3.7.5`, that will also work with this patch.

maxim-belkin: No idea. My understanding was that it still might have something at the end (i.e. `3.7.5.something`) but I may be wrong.

sjackman:
Merged as follows. Thanks for your contribution, Maxim.
```ruby
      kernel_version = `uname -r`.chomp.split("-")[0]
      args << "--enable-kernel=#{kernel_version}" if build.with? "current-kernel"
```

maxim-belkin:
No problem! I like that this solution uses `chomp` and `split`. It looks like, however, it also relies on the dash in the return value of `uname -r`. Can we do something like:
`uname -r`.chomp.split("-")[0].split(".")[0,3].join(".")

sjackman:
I believe (I may be wrong) that kernel versions are either of the format `1.2.3-456.789` or `1.2.3`. Either of those formats works with this code. I'll fix it up if I'm proven wrong.

maxim-belkin:
I'm not worried about the numbers but also about possible suffixes. For example, my Linux appends `.el6.x86_64` suffix. So, if for whatever reason the `uname -r` returns `1.2.3.el6.x86_64` then the formula will fail.

sjackman:
What precisely does `uname -r` return on your system? As long as it includes a hyphen after `1.2.3`, it'll work just fine.

maxim-belkin:
My system returns `2.6.32-504.3.3.el6.x86_64` so it will work on my machine.
I was just trying to take care about a possible situation when the kernel version contains three numbers without a hyphen and is suffixed by something (e.g. `2.6.32.el6.x86_64`) because then the formula is going to fail. However, since neither of us know if there are systems where `uname -r` returns anything like that, we can leave the formula as you wrote it.

PR #796, opened 2016-02-08
mediaconch: fix libxslt dependency

retokromer:


sjackman:
Hi, Reto. For future contributions, please post the exact error message (ideally with a gist) that you're observing without the patch applied.

sjackman:
```
configure: error: libxml2 configuration is not found
```

retokromer:
I will. Thank you for the hint!

sjackman:
The preferred commit message is
`mediaconch: Depends on libxslt for Linuxbrew`

Merged! Thanks for your contribution, Reto.

PR #799, opened 2016-02-09
crystax-ndk: Use linux tarballs

alekseysidorov:


sjackman:
Thanks for your contribution, Aleksey.

PR #800, opened 2016-02-09
Updated mono to latest release: 4.2 SR1 (4.2.2.30)

thedamian:
Just keeping up to date.

Now that ASP CORE 1.0 is compatible with Linux & OSX mono is more useful than ever.


sjackman:
New versions should be submitted upstream to https://github.com/Homebrew/homebrew
See [CONTRIBUTING.md](https://github.com/Linuxbrew/linuxbrew/blob/master/CONTRIBUTING.md)

PR #802, opened 2016-02-11
openssl: add with-ssl2 option

maxim-belkin:
Provide an option to build openssl with ssl2 support if there is a need for one.

### DIFF
--- a/Library/Formula/openssl.rb
+++ b/Library/Formula/openssl.rb
@@ -60,7 +60,7 @@ class Openssl < Formula
   def configure_args; %W[
     --prefix=#{prefix}
     --openssldir=#{openssldir}
-    no-ssl2
+    #{"no-ssl2" if build.without?("ssl2")}
     zlib-dynamic
     shared
     enable-cms


DomT4:
I'm not strongly involved with Linuxbrew and ultimately this is Shaun's call but I am _deeply_ uncomfortable with anything involving SSLv2 going near OpenSSL in 2016. It was killed from Homebrew for a reason.

maxim-belkin:
I agree. But we (the Linux folks) frequently find ourselves leaving in a world that is 10 years behind and some software packages need ssl2 support for whatever reason. So, it would have been nice to be able to turn it on with a simple flag instead of `brew edit ...` "hassle".

sjackman:
I agree with @DomT4, but I'm a pragmatic type. I am also running on a HPC cluster that uses a ten year-old distribution of Linux (CentOS 5), and so I'm sympathetic.

What's your specific use case that you need OpenSSL 2? Perhaps there's a workaround. Have you tried using a [Standalone Installation](https://github.com/Linuxbrew/linuxbrew/wiki/Standalone-Installation) of Linuxbrew?

If there's no workaround that works for you, I'll merge the patch while adding `(not recommended)` to the option descript, and possibly a link to an article explaining why it's not recommended.

DomT4:
My feeling is that Homebrew (and its sister fork) should take the same stance as modern Debian distros do and tell people who need a worryingly insecure OpenSSL to build it themselves, but I tend to take a harder line on this than others.

Library/Formula/openssl.rb
```diff
@@ -57,14 +58,16 @@ def arch_args
...
+  def configure_args
+    args = %W[
+      --prefix=#{prefix}
+      --openssldir=#{openssldir}
+      zlib-dynamic
+      shared
+      enable-cms
+    ]
+    args << "no-ssl2" if build.without?("ssl2")
+    return args
```
sjackman: To keep the patch for Linuxbrew small, please use
```ruby
def configure_args; %W[
…
#{"no-ssl2" if build.without?("ssl2")}
…
]
end
```

maxim-belkin:
Don't get me wrong: I'm not going to use it myself -- I don't need it. But others might. Yes, I use Standalone installation of Linuxbrew. I think adding a `(highly not recommended)` note should suffice.

DomT4:
My concern is that people will do what people tend to do which is to find suggestions online which say "Hey, to solve this problem build OpenSSL with-ssl2 in Linuxbrew" and go "Huh, sure, sounds good" without knowing the potential consequences. We've seen plenty of that sort of behaviour be problematic elsewhere.

sjackman:
If you have no immediate use case, Maxim, I'm going to close this issue for now.

Mikaela:
> My feeling is that Homebrew (and its sister fork) should take the same stance as modern Debian distros do and tell people who need a worryingly insecure OpenSSL to build it themselves, but I tend to take a harder line on this than others.

:+1: and modern distributions have also started compiling OpenSSL without SSLv3 support and I think Linuxbrew should also follow with that unless it's already done.

DomT4:
That should land in Homebrew when OpenSSL 1.1.0 does.

I decided to bundle all the API breaks (OpenSSL 1.1.0, SSLv3 removal, zlib-support removal) into an alarmingly big PR when the time comes because each will require almost everything to be rebuilt, so doing them separately would get annoying for people quickly :smile:.

PR #806, opened 2016-02-12
A patch block depends on gpatch for Linuxbrew

sjackman:
@DoomHammer @KomodoTech


sjackman:
https://travis-ci.org/Linuxbrew/linuxbrew/builds/108714964#L1011
```
 * Formula name conflicts with homebrew/dupes/gpatch
```

sjackman:
`zlib` has a patch for `clang`, and `binutils` has a patch for Mac OS 10.6. These two patches should be disabled to simplify the dependency chain for Linuxbrew.

sjackman:
Merged.

PR #808, opened 2016-02-14
glibc: fix for chmod in postinstall

maxim-belkin:
When I try to use current glibc formula, it fails with the following error message: 
```
$ brew postinstall glibc
Error: can't convert String into Integer
```

### DIFF
--- a/Library/Formula/glibc.rb
+++ b/Library/Formula/glibc.rb
@@ -36,7 +36,7 @@ class Glibc < Formula
 
   def post_install
     # Fix permissions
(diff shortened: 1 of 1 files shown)


sjackman:
Thanks for the fix. Which version of Ruby are you using, Maxim?

maxim-belkin:
`irb --version` returns `0.9.5`

sjackman:
Ruby 1.8 has only integer mode chmod. Ruby 1.9 introduced string mode.

sjackman:
Merged. Thanks for the fix, Maxim.

PR #828, opened 2016-02-21
ansible: Fix for Linuxbrew

hartzell:
### All Submissions:

- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew/pulls) for the same update/change?

---

The Babel package died looking for pytz.  If I move the pytz
resource ahead of Babel's then it works.

### DIFF
--- a/Library/Formula/ansible.rb
+++ b/Library/Formula/ansible.rb
@@ -22,6 +22,7 @@ class Ansible < Formula
   depends_on :python if MacOS.version <= :snow_leopard
   depends_on "libyaml"
   depends_on "openssl"
+  depends_on "libffi" unless OS.mac?
 
   #
   # ansible (core dependencies)
@@ -215,6 +216,11 @@ class Ansible < Formula
   #
   # pyrax (for Rackspace support)
   #
+  resource "pytz" do
+    url "https://pypi.python.org/packages/source/p/pytz/pytz-2015.7.tar.bz2"
+    sha256 "fbd26746772c24cb93c8b97cbdad5cb9e46c86bbdb1b9d8a743ee00e2fb1fc5d"
+  end
+
   resource "Babel" do
     url "https://pypi.python.org/packages/source/B/Babel/Babel-2.2.0.tar.gz"
     sha256 "d8cb4c0e78148aee89560f9fe21587aa57739c975bb89ff66b1e842cc697428f"
@@ -334,11 +340,6 @@ class Ansible < Formula
   resource "python-novaclient" do
     url "https://pypi.python.org/packages/source/p/python-novaclient/python-novaclient-3.2.0.tar.gz"
     sha256 "b00a70ba72b068adb6aaca19b14ed4bde9c950ef4f24234d095f9f9985073179"
-  end
-
-  resource "pytz" do
-    url "https://pypi.python.org/packages/source/p/pytz/pytz-2015.7.tar.bz2"
-    sha256 "fbd26746772c24cb93c8b97cbdad5cb9e46c86bbdb1b9d8a743ee00e2fb1fc5d"
   end
 
   resource "rackspace-auth-openstack" do


hartzell:
I'd like to keep plugging away at this but could use some guidance.

I've been trying to get things going on real live linux boxes, but that might be too confusing (I have no control of what's installed where when...).

I'm working on getting the linuxbrew docker images/containers up and running.

sjackman:
Thanks for your contribution, George. I've added the constant `OS::NAME` in ce0f667 just now to make your patch easier. Its value is either `darwin` on Mac OS or `linux` on Linux.

Library/Formula/ansible.rb
```diff
@@ -22,6 +22,7 @@ class Ansible < Formula
   depends_on :python if MacOS.version <= :snow_leopard
   depends_on "libyaml"
   depends_on "openssl"
+  depends_on "libffi" if OS.linux?
```
sjackman: Please change this line to
```ruby
depends_on "libffi" unless OS.mac?
```

sjackman: Also please report the error message that you see without this patch. `brew gist-logs ansible` is helpful.

Library/Formula/ansible.rb
```diff
@@ -331,11 +337,6 @@ class Ansible < Formula
     sha256 "b00a70ba72b068adb6aaca19b14ed4bde9c950ef4f24234d095f9f9985073179"
   end
 
-  resource "pytz" do
-    url "https://pypi.python.org/packages/source/p/pytz/pytz-2015.7.tar.bz2"
-    sha256 "fbd26746772c24cb93c8b97cbdad5cb9e46c86bbdb1b9d8a743ee00e2fb1fc5d"
-  end
-
```
sjackman: Please remove this chunk of the patch if it has no purpose.

sjackman:
Are you still working on this patch, George?

Library/Formula/ansible.rb
hartzell: I'm not sure which hunk you'd like me to remove.  The patch moves pytz to a different location in the file, the first hunk inserts it into the new location and the second hunk deletes it from the old.

Library/Formula/ansible.rb
hartzell: https://gist.github.com/efdb86973be7483a4592

hartzell: Digging a bit more.  I've made the change to use `unless OS.mac?`.  I'm still having trouble.

I think that the problem is that the *libffi* formula does not install *ffi.h` into the include directory, so building cffi fails.

*Should* the *libffi* formula copy/link/... *ffi.h*?

sjackman: The `ffi` formula installs `ffi.h` in a subdirectory, which seems to be the default behaviour of `libffi`.
```
$ brew list libffi |grep ffi.h
/gsc/btl/linuxbrew/Cellar/libffi/3.0.13/lib/libffi-3.0.13/include/ffi.h
$ ls `brew --prefix libffi`/lib/libffi-3.0.13/include
ffi.h  ffitarget.h
```

Library/Formula/ansible.rb
sjackman: Ah, I see. I didn't realize that the Python resources are installed in the order that they're listed in this file by magics later on. Sorry for my confusing comment.

Library/Formula/ansible.rb
sjackman: > I've made the change to use unless OS.mac?

I don't see this change yet. You can update your PR with `brew push --force`

sjackman:
Please squash the two commits to a single commit, and use set the first line of the log message to
`ansible: Fix for Linuxbrew`
The rest of the log message can be whatever you want.
To squash the two commits run
```sh
git checkout ansible-on-linux
brew rebase -i HEAD~2
git push --force hartzell ansible-on-linux
```
I'll merge this PR after the squash.

hartzell:
Squashed and pushed.

By the way, the squash/rebase instructions were *very* helpful.  I probably would have screwed around a while and possibly blown it.  Three lines from you made the commit happen easily.  (Although the second line is `git`, not `brew`...).

sjackman:
> By the way, the squash/rebase instructions were very helpful.

Glad to hear it! I'll add those instructions to the `CONTRIBUTING.md`.

> (Although the second line is git, not brew...).

Fixed above, for posterity. Thanks.

sjackman:
Merged. Thanks for your contribution, George, and for sticking with it!

Issue #829, opened 2016-02-22
coreutils: linking error. Conflicting dependency: util-linux


maxim-belkin:
See pull request #830 first.
The problem is as follows: when `coreutils` is built from source, linuxbrew installs `util-linux` as a dependency of `wget`. The relevant content of **`wget.rb`**:

`depends_on "util-linux" if OS.linux? # for libuuid`
Suggested change:
`depends_on "libuuid" if OS.linux? # for libuuid`

 `Util-linux` conflicts with `coreutils` because of a common for them program `kill`. A relevant web-page: [link](http://lists.busybox.net/pipermail/buildroot/2014-July/100759.html)

At the same time, neither `wget` nor `util-linux` are reported as a dependency of `coreutils` by `brew deps coreutils` or `brew deps --tree coreutils` or `brew deps --tree --installed`. So, not sure what's going on there... But my main message of the report is faulty dependency tree for `coreutils`.

1. First failed attempt to install `coreutils --HEAD`

        $ brew install coreutils --HEAD
        ==> Installing dependencies for coreutils: homebrew/dupes/m4, autoconf, automake, bison, texinfo, util-linux, wget
        ==> Installing coreutils dependency: homebrew/dupes/m4
        ==> Downloading http://ftpmirror.gnu.org/m4/m4-1.4.17.tar.xz
        ==> Downloading from http://mirror.thecodefactory.org/gnu/m4/m4-1.4.17.tar.xz
        ######################################################################## 100.0%
        ==> ./configure --prefix=[Linuxbrew-folder]/Cellar/m4/1.4.17
        ==> make
        ==> make install
        [Linuxbrew-folder]/Cellar/m4/1.4.17: 12 files, 762.7K, built in 43 seconds
        ==> Installing coreutils dependency: autoconf
        ==> Downloading http://ftpmirror.gnu.org/autoconf/autoconf-2.69.tar.gz
        ==> Downloading from http://mirror.nexcess.net/gnu/autoconf/autoconf-2.69.tar.gz
        ######################################################################## 100.0%
        ==> ./configure --prefix=[Linuxbrew-folder]/Cellar/autoconf/2.69 --with-lispdir=[Linuxbrew-folder]/
        ==> make install
        ==> Caveats
        
        Emacs Lisp files have been installed to:
          [Linuxbrew-folder]/share/emacs/site-lisp/autoconf
        ==> Summary
[...]

PR #830, opened 2016-02-22
fixes for coreutils and wget

maxim-belkin:
**`coreutils`**:
1. Changed the url used for building a HEAD version. Current one seems to be valid but non-functional. 
2. Added `gperf` as a dependency because it is required during the bootstrapping phase

**`wget`**:
1. Change dependency on `util-linux` to `libuuid` because `wget` is a build-dependency of `coreutils` which has minor conflict with `util-linux`.

### DIFF
--- a/Library/Formula/coreutils.rb
+++ b/Library/Formula/coreutils.rb
@@ -29,6 +29,7 @@ class Coreutils < Formula
     depends_on "texinfo" => :build
     depends_on "xz" => :build
     depends_on "wget" => :build
+    depends_on "homebrew/dupes/gperf" => :build unless OS.mac?
   end
 
   depends_on "gmp" => :optional
--- a/Library/Formula/coreutils.rb
+++ b/Library/Formula/coreutils.rb
@@ -20,7 +20,7 @@ class Coreutils < Formula
   conflicts_with "aardvark_shell_utils", :because => "both install `realpath` binaries"
 
   head do
-    url "git://git.sv.gnu.org/coreutils"
+    url "http://git.savannah.gnu.org/r/coreutils.git"
 
     depends_on "autoconf" => :build
     depends_on "automake" => :build


Library/Formula/coreutils.rb
```diff
@@ -20,7 +20,7 @@ class Coreutils < Formula
   conflicts_with "aardvark_shell_utils", :because => "both install `realpath` binaries"
 
   head do
-    url "git://git.sv.gnu.org/coreutils"
+    url "http://git.savannah.gnu.org/r/coreutils.git"
```
sjackman: Please either remove this chunk or send it upstream to Homebrew, as it's not specific to Linux.

Library/Formula/coreutils.rb
```diff
@@ -29,6 +29,7 @@ class Coreutils < Formula
     depends_on "texinfo" => :build
     depends_on "xz" => :build
     depends_on "wget" => :build
+    depends_on "gperf" => :build
```
sjackman: Please include a log of the failed build without your patch using `brew gist-logs coreutils`.

Library/Formula/wget.rb
```diff
@@ -37,7 +37,7 @@ class Wget < Formula
   depends_on "pcre" => :optional
   depends_on "libmetalink" => :optional
   depends_on "gpgme" => :optional
-  depends_on "util-linux" if OS.linux? # for libuuid
+  depends_on "libuuid" if OS.linux? # for libuuid
```
sjackman: Please remove this chunk. See comment in #829.

Library/Formula/coreutils.rb
maxim-belkin: ok

Library/Formula/coreutils.rb
maxim-belkin: https://gist.github.com/maxim-belkin/1600f4414a8d0c9f4c99

sjackman:
```
./bootstrap: line 221: gperf: command not found
./bootstrap: Error: 'gperf' not found

./bootstrap: See README-prereq for how to get the prerequisite programs
```
https://gist.github.com/maxim-belkin/1600f4414a8d0c9f4c99#file-01-bootstrap-L5-L8

sjackman: Please change this line to
``ruby
depends_on "homebrew/dupes/gperf" => :build unless OS.mac?
```

Comment (sjackman) [inline Library/Formula/coreutils.rb]: I cannot reproduce this error. Does this issue affect only `HEAD`?

Comment (sjackman) [inline Library/Formula/coreutils.rb]: Yes. `system "./bootstrap" if build.head?`
```ruby
depends_on "homebrew/dupes/gperf" => :build if build.head? && !OS.mac?
```

maxim-belkin: But this part is inside of a `head` section. Shouldn't it be just `depends_on "homebrew/dupes/gperf" => :build && !OS.mac?`

sjackman: My mistake. Yes, that's correct.
```ruby
depends_on "homebrew/dupes/gperf" => :build unless OS.mac?
```

sjackman:
Merged the `gperf` dependency. If you send the `url` change upstream to Homebrew, it'll get merged into Linuxbrew at the next merge. Thanks for your contribution, Maxim.

PR #840, opened 2016-02-24
serf: fix for Linux

maxim-belkin:
Had to do this change to make `serf` compile.

### DIFF
--- a/Library/Formula/serf.rb
+++ b/Library/Formula/serf.rb
@@ -106,7 +106,7 @@ class Serf < Formula
     ENV["GOPATH"] = gopath
     arch = MacOS.prefer_64_bit? ? "amd64" : "386"
     ENV["XC_ARCH"] = arch
-    ENV["XC_OS"] = "darwin"
+    ENV["XC_OS"] = OS::NAME
     ENV.prepend_create_path "PATH", gopath/"bin"
 
     Language::Go.stage_deps resources, gopath/"src"


sjackman:
I think this `serf` formula is different than the `serf` mentioned in `subversion`. This one is https://serfdom.io/
The one in `subversion` is http://serf.apache.org/

Library/Formula/serf.rb
```diff
@@ -106,7 +106,11 @@ def install
     ENV["GOPATH"] = gopath
     arch = MacOS.prefer_64_bit? ? "amd64" : "386"
     ENV["XC_ARCH"] = arch
-    ENV["XC_OS"] = "darwin"
+    if OS.mac?
+      ENV["XC_OS"] = "darwin"
+    else
+      ENV["XC_OS"] = "linux"
+    end
```
sjackman: Please change this to
```ruby
ENV["XC_OS"] = OS::NAME
```

sjackman:
Thanks for the fix, Maxim. I'll be happy to merge it after the suggested change.

maxim-belkin:
Sorry for the delay. I've updated the commit.

sjackman:
Unrelated error when installing `curl`
https://travis-ci.org/Linuxbrew/linuxbrew/builds/111801055#L1229
```
/usr/bin/perl ./zsh.pl ../src/curl > _curl
/tmp/curl20160225-30127-6bwhx2/curl-7.47.1/src/.libs/lt-curl: error while loading shared libraries: liblber-2.4.so.2: cannot open shared object file: No such file or directory
```
`liblber` is related to LDAP. https://packages.debian.org/sid/amd64/libldap-2.4-2/filelist

sjackman:
No worries. Merged. Thanks for your contribution, Maxim.

PR #844, opened 2016-02-25
subversion: Depends on zlib

eiennohito:
Otherwise the build was failing on CentOS 6.7

 https://gist.github.com/anonymous/1613a9b672b3215546c4

### DIFF
--- a/Library/Formula/subversion.rb
+++ b/Library/Formula/subversion.rb
@@ -49,6 +49,7 @@ class Subversion < Formula
   unless OS.mac?
     depends_on "apr"
     depends_on "apr-util"
+    depends_on "zlib"
   end
 
   # Other optional dependencies
@@ -123,7 +124,7 @@ class Subversion < Formula
     args = %W[
       --disable-debug
       --prefix=#{prefix}
-      --with-zlib=/usr
+      #{"--with-zlib=/usr" if OS.mac?}
       --with-sqlite=#{Formula["sqlite"].opt_prefix}
       #{"--with-serf=#{serf_prefix}" if build.with? "serf"}
       --disable-mod-activation


Library/Formula/subversion.rb
```diff
@@ -138,7 +138,9 @@ def install
     if OS.mac? && MacOS::CLT.installed?
       args << "--with-apr=/usr"
       args << "--with-apr-util=/usr"
+      args << "--with-zlib=/usr"
     else
+      args << "--with-zlib=#{Formula["zlib"].opt_prefix}"
```
sjackman: This chunk isn't quite right for Mac, because zlib on Mac doesn't require that CLT be installed.
Please remove this chunk (both lines).

Library/Formula/subversion.rb
```diff
@@ -123,7 +124,6 @@ def install
     args = %W[
       --disable-debug
       --prefix=#{prefix}
-      --with-zlib=/usr
```
sjackman: Please change this chunk to
```ruby
#{"--with-zlib=/usr" if OS.mac?}
```

sjackman:
In the gist that you posted, I couldn't find an error related to `zlib`. Just this error:
https://gist.github.com/anonymous/1613a9b672b3215546c4#file-02-make-L27
```
/usr/include/stdlib.h:521:1: error: C++ style comments are not allowed in ISO C90
 // XXX There should be a macro to signal with C++ revision is used.
```

eiennohito:
It manifested when using `/usr` as include directory for zlib.
Brewed gcc does not have `/usr/include` in the include path by default and everything goes wild when it is added.

sjackman:
> It manifested when using /usr as include directory for zlib.

Right. Of course. I understand. Yes, good patch. Please fix up my comments, and I'll merge it.

eiennohito:
Okay, fixed as per comments

sjackman:
https://travis-ci.org/Linuxbrew/linuxbrew/builds/111967566
```
Testing subversion
==> svnadmin create test
/home/travis/build/Linuxbrew/linuxbrew/Cellar/subversion/1.9.3/bin/svnadmin: error while loading shared libraries: libapr-1.so.0: cannot open shared object file: No such file or directory
Error: subversion: failed
Failed executing: /home/travis/build/Linuxbrew/linuxbrew/Cellar/subversion/1.9.3/bin/svnadmin create test
```

sjackman:
Strange error from Travis, unrelated to this patch.

sjackman:
Merged. Thanks for your contribution, Arseny!

Issue #412, opened 2015-05-21
rust: cargo fails to build on Debian 7.4


pramodhegde:
```
%brew install rust
==> Downloading https://static.rust-lang.org/dist/rustc-1.0.0-src.tar.gz
######################################################################## 100.0%
==> ./configure --prefix=/home/user/.linuxbrew/Cellar/rust/1.0.0 --release-channel=stable
==> make
==> make install
==> Cloning https://github.com/rust-lang/cargo.git
Updating /home/user/.cache/Homebrew/rust--cargo--git
Synchronizing submodule url for 'test/rust-installer-v1'
Synchronizing submodule url for 'test/rust-installer-v2'
Synchronizing submodule url for 'test/rust-installer-v1'
==> Checking out revision a48358155c90467ed9c897930dd0da4614605dac
==> Downloading https://static-rust-lang-org.s3.amazonaws.com/dist/2015-05-14/rustc-nightly-x86_64-apple-darwin.tar.gz
######################################################################## 100.0%
==> ./install.sh --prefix=/tmp/rust--cargo20150521-24860-1hswv02/rustc
==> Downloading https://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2015-05-14/cargo-nightly-x86_64-apple-darwin.tar.gz
######################################################################## 100.0%
==> ./install.sh --prefix=/tmp/rust--cargo20150521-24860-1hswv02/target/snapshot/cargo
==> ./configure --prefix=/home/user/.linuxbrew/Cellar/rust/1.0.0 --local-rust-root=/tmp/rust--cargo20150521-24860-1hswv02/rustc
configure: 
configure: CFG_PREFIX           := /home/user/.linuxbrew/C ...
configure: CFG_LOCAL_RUST_ROOT  := /tmp/rust--cargo20150521-24860-1hsw ...
/tmp/rust--cargo20150521-24860-1hswv02/rustc/bin/rustc: 7: /tmp/rust--cargo20150521-24860-1hswv02/rustc/bin/rustc: Syntax error: Unterminated quoted string
configure: error: failed to run rustc at: /tmp/rust--cargo20150521-24860-1hswv02/rustc
```

Gist: https://gist.github.com/anonymous/f01adc7faf51bfac4b87

sjackman:
Hi, Pramod. I'm afraid that I don't have the time to troubleshoot individual formula. If you're able to find a fix to this issue, please submit a PR.

[...]

PR #846, opened 2016-02-25
rust: Fix Rust 1.6.0 build

palad1:
Fix bootstraping of Cargo by using a newer snapshot of Cargo compatible with the current version of crates.io

### DIFF
--- a/Library/Formula/rust.rb
+++ b/Library/Formula/rust.rb
@@ -86,10 +86,6 @@ class Rust < Formula
       end
 
       system "./configure", "--prefix=#{prefix}", "--local-rust-root=#{prefix}", "--enable-optimize"
-      
-      # Snapshots.txt for 1.6.0 are NOT compatible with crates.io
-      system 
-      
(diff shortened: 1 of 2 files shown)


sjackman:
What error message without this patch are you trying to fix? Please report `brew gist-logs rust`.

Can you test on Mac, and does this issue also affect Mac OS? If so, please send this patch upstream to Homebrew/homebrew.

rwhogg:
Looks potentially related to #412.

rwhogg:
@palad1 new versions of Rust/Cargo were committed [here](https://github.com/Linuxbrew/linuxbrew/commit/f939575005011d0681227aa530552d264edd933a#diff-5228b5cf14fdcdd1f7e2c1fd8091ad86).

Does the problem still occur for you?

sjackman:
I believe the `cargo` resource has been bumped to `0.9.0` by upstream Homebrew and merged into Linuxbrew, resolving this issue.

PR #848, opened 2016-02-26
cmake: Depends on bzip2

zbeekman:
### All Submissions:

- [X] Have you followed the guidelines in our [Contributing](https://github.com/Linuxbrew/linuxbrew/blob/master/.github/CONTRIBUTING.md) document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Linuxbrew/linuxbrew/pulls) for the same update/change?
- [ ] Have you included a log of the failed build without your patch using `brew gist-logs <formula>`?
    - Build doesn't fail because it downloads bottle
    - Build from source takes FOREVER for CMake. I'll try to upload a log in the morning
- [X] Does your submission pass
- [ ] Have you built your formula locally prior to submission with `brew install <formula>`?
  - I'll try overnight. Problems may persist with other dependency issues. Specifically the default Sphinx dependency is a real PITA

### DIFF
--- a/Library/Formula/cmake.rb
+++ b/Library/Formula/cmake.rb
@@ -3,6 +3,7 @@ class Cmake < Formula
   homepage "https://www.cmake.org/"
   url "https://cmake.org/files/v3.4/cmake-3.4.3.tar.gz"
   sha256 "b73f8c1029611df7ed81796bf5ca8ba0ef41c6761132340c73ffe42704f980fa"
+  revision 1
   head "https://cmake.org/cmake.git"
 
   bottle do
@@ -18,6 +19,7 @@ class Cmake < Formula
 
   depends_on "sphinx-doc" => :build if build.with? "docs"
   depends_on "curl" unless OS.mac?
+  depends_on "bzip2" unless OS.mac?
 
   # The `with-qt` GUI option was removed due to circular dependencies if
   # CMake is built with Qt support and Qt is built with MySQL support as MySQL uses CMake.


zbeekman:
An additional comment: The travis-ci environment is not as virginal as a bare ubuntu installation, so that is likely why the bzip2 dependency was not detected.

zbeekman:
Seems like there is a bigger issue that is breaking the travis build...

zbeekman:
Right now on my local (vagrant) machine, `brew install --only-dependencies cmake` fails for curl: https://gist.github.com/3355ded791bc80a3e359

zbeekman:
It also fails with sphinx-doc:
https://gist.github.com/a567e0bb1a76107b9aa0

Library/Formula/cmake.rb
```diff
@@ -3,6 +3,7 @@ class Cmake < Formula
   homepage "https://www.cmake.org/"
   url "https://cmake.org/files/v3.4/cmake-3.4.3.tar.gz"
   sha256 "b73f8c1029611df7ed81796bf5ca8ba0ef41c6761132340c73ffe42704f980fa"
+  revision 1
```
sjackman: I haven't been bumping `revision` for Linuxbrew. I keep the revision numbers in sync with Homebrew.

sjackman:
Merged. Thanks for your contribution, Izaak.

sjackman:
```
/usr/bin/perl ./zsh.pl ../src/curl > _curl
/tmp/curl20160226-32688-7lw21u/curl-7.47.1/src/.libs/lt-curl: error while loading shared libraries: liblber-2.4.so.2: cannot open shared object file: No such file or directory
```
https://gist.github.com/anonymous/3355ded791bc80a3e359#file-02-make-L473

I've seen this error as well on some systems. `liblber-2.4.so.2` is provided by `homebrew/dupes/openldap`. I think this happens when the host system has `openldap` installed, but Homebrew does not.

Library/Formula/cmake.rb
zbeekman: OK, I'll try to remember this for my next PR

zbeekman:
Any thoughts on the [sphinx-doc failure](https://gist.github.com/a567e0bb1a76107b9aa0)?

zbeekman:
> the option is to make openldap a dependency of curl.

Would you accept a PR to make openldap a dependency of curl?

sjackman:
> Would you accept a PR to make openldap a dependency of curl?

There's some interaction with the host system, which I'd rather avoid. `ldap` isn't a hard dependency of `curl`. I'd rather make it an `:optional` dependency. Is there a `configure --without-ldap` option for `curl`?

sjackman:
> Any thoughts on the sphinx-doc failure?

https://gist.github.com/anonymous/a567e0bb1a76107b9aa0#file-01-python-L17
`ImportError: No module named setuptools`

Instal your host system's Python package for `setuptools`, or `brew install python`.

sjackman:
The `setuptools` issue has come up about eleven times before. A `brew doctor` test or some other solution seems in order.

zbeekman:
`pypy-setuptools` or `python-setuptools`? (Or does it matter?)

sjackman:
I don't know definitevely the answer to that question. I would go with `sudo apt-get install python-setuptools`.

PR #851, opened 2016-02-26
curl: Depends on openldap for Linuxbrew

zbeekman:
Tested with hashicorp/precise64 image

### DIFF
--- a/Library/Formula/curl.rb
+++ b/Library/Formula/curl.rb
@@ -21,7 +21,6 @@ class Curl < Formula
   option "with-libmetalink", "Build with libmetalink support."
   option "with-libressl", "Build with LibreSSL instead of Secure Transport or OpenSSL"
   option "with-nghttp2", "Build with HTTP/2 support (requires OpenSSL or LibreSSL)"
-  option "without-ldap", "Build without LDAP support"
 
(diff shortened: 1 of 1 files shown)


sjackman:
Please change this to
```ruby
depends_on "Homebrew/dupes/openldap" => :optional
```
Please remove
```ruby
  option "without-ldap", "Build without LDAP support"
```

zbeekman:
> Please change this to
> 
> depends_on "Homebrew/dupes/openldap" => :optional unless OS.mac?
> Please remove
> 
>   option "without-ldap", "Build without LDAP support"

Good suggestions. I'm still getting used to the succinct power or Ruby.

zbeekman:
Do changes to formulae for Linuxbrew get contributed back upstream, or will they at some point? Just curious because removing the `option "without-ldap"` will remove this option completely from the OS.mac builds.

sjackman:
The option `option "without-ldap"` is in fact specific to the Linuxbrew formula. Typically Linuxbrew patches do not get merged back upstream in Homebrew.

sjackman:
Merged. Thanks for your contribution, Izaak!

zbeekman:
:tada: Thanks again for the awesome project

PR #856, opened 2016-02-27
gnu-tar: Avoid patch on linux

yesimon:
The patch includes changes to `Makefile.in`. Therefore the `./configure` script realizes that the file has changed and needs to regenerate `Makefile`, which hence requires the autotools suite. However, the autotools suite has set the `am__api_version` to be 1.14, while the newest brew version is 1.15. Since the configure script explicitly requires the versioned `automake-1.14` binary, this doesn't compile with brew autotools. 

Obviously this can be circumvented if we don't patch, but I'm not sure if this is the best way to fix this problem. The alternative would be to require autotools for build, and run `autoreconf -fi` before configure, but this may be considered heavy handed. It would be great if there were some way to link the usage of autotools if certain files were patched.

### DIFF
--- a/Library/Formula/gnu-tar.rb
+++ b/Library/Formula/gnu-tar.rb
@@ -19,7 +19,7 @@ class GnuTar < Formula
   patch do
     url "https://gist.githubusercontent.com/mistydemeo/10fbae8b8441359ba86d/raw/e5c183b72036821856f9e82b46fba6185e10e8b9/gnutar-configure-xattrs.patch"
     sha256 "f2e56bb8afd1c641a7e5b81e35fdbf36b6fb66434b1e35caa8b55196b30c3ad9"
-  end
+  end if OS.mac?
 
   def install
     # Work around unremovable, nested dirs bug that affects lots of


Library/Formula/gnu-tar.rb
```diff
@@ -14,11 +14,13 @@ class GnuTar < Formula
...
 
-  # Fix for xattrs bug causing build failures on OS X:
-  # https://lists.gnu.org/archive/html/bug-tar/2014-08/msg00001.html
-  patch do
-    url "https://gist.githubusercontent.com/mistydemeo/10fbae8b8441359ba86d/raw/e5c183b72036821856f9e82b46fba6185e10e8b9/gnutar-configure-xattrs.patch"
-    sha256 "f2e56bb8afd1c641a7e5b81e35fdbf36b6fb66434b1e35caa8b55196b30c3ad9"
+  if OS.mac?
+    # Fix for xattrs bug causing build failures on OS X:
+    # https://lists.gnu.org/archive/html/bug-tar/2014-08/msg00001.html
+    patch do
```
sjackman: Please change this to
```ruby
patch do
…
end if OS.mac?
```

sjackman:
It's not patching `Makefile.in` that's the problem, but patching `src/Makefile.am` is causing the troubles.

sjackman:
Thanks for your fix, Simon. Fix the above suggestion, and I'll merge it. Cheers.

yesimon:
It's not clear to me why travis is failing but I have updated it.

yesimon:
Sorry, I do feel bad about this change because it's papering over the problem of the interaction between patches and autotools. I think this is probably still broken on OS X, but this formula probably isn't very popular. Perhaps I should start an issue in the homebrew issue tracker?

sjackman:
> Perhaps I should start an issue in the homebrew issue tracker?

If you can reproduce this issue on Mac, then yes, please do. In principle, it should also affect Mac.

sjackman:
> Sorry, I do feel bad about this change because it's papering over the problem of the interaction between patches and autotools.

It's not ideal, but if this patch is only needed for Mac, then it's a reasonable, pragmatic solution.

sjackman:
```
Failed to execute: /home/travis/build/Linuxbrew/linuxbrew/Cellar/gnu-tar/1.28/bin/gtar
Error: gnu-tar: failed
```
https://travis-ci.org/Linuxbrew/linuxbrew/builds/112362026#L562

sjackman:
I've fixed the `test` failure with 84209b7

sjackman:
Merged. Thanks for your contribution, Simon!

PR #857, opened 2016-02-28
wxmac: fix build for linux

### LINKED ISSUE #512: erlang: Install LFE got Error when installing dependeny wxmac
_From @guawoo on August 14, 2015 2:9_

```
==> ./configure --prefix=/home/guawoo/.linuxbrew/Cellar/wxmac/3.0.2 --enable-unicode --enable-std_string --ena
checking for ar... ar
checking for architectures to use in universal binary... 
checking if C compiler (/usr/bin/gcc-4.9 -mmacosx-version-min=0) works with SDK/version options... configure: error: in `/tmp/wxmac20150814-96681-uy7369/wxWidgets-3.0.2':
configure: error: no.  Try a different SDK

READ THIS: https://github.com/Homebrew/linuxbrew/blob/master/share/doc/homebrew/Troubleshooting.md#troubleshooting
```

_Copied from original issue: Homebrew/homebrew#42925_

rwhogg:
See https://gist.github.com/rwhogg/33906c99f0335a9d6705

Closes #512

Message:
> checking if C compiler (/home/bob/.linuxbrew/bin/gcc-5 -mmacosx-version-min=0) works with SDK/version options... configure: error: in `/home/bob/tmp/wxmac20160227-6760-1wyythr/wxWidgets-3.0.2':
configure: error: no.  Try a different SDK

### DIFF
--- a/Library/Formula/wxmac.rb
+++ b/Library/Formula/wxmac.rb
@@ -19,6 +19,7 @@ class Wxmac < Formula
   depends_on "jpeg"
   depends_on "libpng"
   depends_on "libtiff"
+  depends_on "gtk+" unless OS.mac?
 
   # Various fixes related to Yosemite. Revisit in next stable release.
   # Please keep an eye on http://trac.wxwidgets.org/ticket/16329 as well
@@ -34,7 +35,6 @@ class Wxmac < Formula
       "--enable-std_string",
       "--enable-display",
       "--with-opengl",
-      "--with-osx_cocoa",
       "--with-libjpeg",
       "--with-libtiff",
       # Otherwise, even in superenv, the internal libtiff can pick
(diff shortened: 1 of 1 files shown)


rwhogg:
Note: I didn't change the name, even though it no longer really makes sense. If you agree, I can file an issue upstream to rename it.

rwhogg:
The Travis build failed. Looks like it's because of a missing dependency on GTK+. I'll add that.

sjackman:
> Note: I didn't change the name, even though it no longer really makes sense. If you agree, I can file an issue upstream to rename it.

I agree that the formula would be better named `wxwidgets`. I can't say whether the Homebrew maintainers will agree, but you can ask.

sjackman:
Merged. Thanks for you contribution, Bob!

rwhogg:
There's already an alias, so I guess renaming isn't necessary.

As a side note, they also requested that I not create issues of this sort upstream anymore. They'd prefer that these generally get fixed downstream instead.

PR #869, opened 2016-02-29
gcc: Add brew env LDFLAGS to boot-ldflags on non-standalone

yesimon:
My system had trouble finding isl even with the --with-isl configure
options. Adding the brew LDFLAGS to boot-ldflags is probably more
desired in all cases because that will also set the rpath on our gcc
binaries to the linuxbrew lib path.

Note that by default `--with-boot-ldflags` has the `-static-libgcc` and `-static-libstdc++` set if empty so they need to be repeated.

### DIFF
--- a/Library/Formula/gcc.rb
+++ b/Library/Formula/gcc.rb
@@ -118,7 +118,6 @@ class Gcc < Formula
         "--with-native-system-header-dir=#{HOMEBREW_PREFIX}/include",
         "--with-local-prefix=#{HOMEBREW_PREFIX}/local",
         "--with-build-time-tools=#{binutils}",
-        "--with-boot-ldflags=-static-libstdc++ -static-libgcc #{ENV["LDFLAGS"]}",
       ]
       # Set the search path for glibc libraries and objects.
       ENV["LIBRARY_PATH"] = Formula["glibc"].lib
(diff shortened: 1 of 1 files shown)


sjackman:
Can you please include a log of the failed build without your patch using `brew gist-logs gcc`?

sjackman:
The Travis failure is due to an unrelated workaround hack for a buggy `glibc`.
```
26.00s==> brew install --verbose --build-bottle gcc
==> Downloading http://ftpmirror.gnu.org/gcc/gcc-5.3.0/gcc-5.3.0.tar.bz2
Already downloaded: /home/travis/.cache/Homebrew/gcc-5.3.0.tar.bz2
==> Verifying gcc-5.3.0.tar.bz2 checksum
tar xf /home/travis/.cache/Homebrew/gcc-5.3.0.tar.bz2
Error: File exists - /home/travis/build/Linuxbrew/linuxbrew/Cellar/gcc/5.3.0/lib
==> FAILED
```

sjackman:
I was able to reproduce this error:
```
/home/sjackman/.linuxbrew/Cellar/gcc/5.3.0/x86_64-unknown-linux-gnu/sys-include 
   -o conftest -g -O2 -Os -w -pipe -march=core2   conftest.c  >&5
/var/tmp/sjackman/gcc20160229-27179-1blm7mu/gcc-5.3.0/build/./gcc/cc1: error while loading shared libraries: libisl.so.15: cannot open shared object file: No such file or directory
```

sjackman:
Merged. Thanks for your contribution, Simon!

PR #872, opened 2016-03-01
scons: add homebrew default path to scons PATH variable

### LINKED ISSUE #845: scons ignores PATH environment variable (mongodb does not build on CentOS 6.7)
I am trying to setup Linuxbrew environment on CentOS 6.7

Several formulas (subversions's serf, mongodb) are using scons for the build and it seems that 
it uses own definition of PATH to find various tools as a feature.
http://scons.org/doc/1.1.0/HTML/scons-user/x1673.html

This results in very funny errors:
- Output: https://gist.github.com/eiennohito/4406be3e901cd6084c4c
- config.log: https://gist.github.com/anonymous/6940d56dbf847997c5bc#file-config-log-L425

It seems that because of `PATH` issues, system `ld` was used instead of brewed gcc one.

Patching `SConstruct` file seem hacky, more common solution would be better I believe.
If you have any ideas, please share them.

Default scons `PATH` is hardcoded:
https://github.com/azatoth/scons/blob/master/src/engine/SCons/Platform/posix.py#L233
Maybe we should patch it?

eiennohito:
should fix #845 per discussion in the issue

### DIFF
--- a/Library/Formula/scons.rb
+++ b/Library/Formula/scons.rb
@@ -13,6 +13,11 @@ class Scons < Formula
 
   def install
     man1.install gzip("scons-time.1", "scons.1", "sconsign.1")
+    inreplace(
+      "engine/SCons/Platform/posix.py",
+      "env['ENV']['PATH']    = '/usr/local/bin:/opt/bin:/bin:/usr/bin'",
+      "env['ENV']['PATH']    = '#{HOMEBREW_PREFIX/'bin'}:/usr/local/bin:/opt/bin:/bin:/usr/bin'"
+    )
     system "/usr/bin/python", "setup.py", "install",
              "--prefix=#{prefix}",
              "--standalone-lib",


Library/Formula/scons.rb
```diff
@@ -13,6 +13,11 @@ class Scons < Formula
 
   def install
     man1.install gzip("scons-time.1", "scons.1", "sconsign.1")
+    inreplace(
+      "engine/SCons/Platform/posix.py",
+      "env['ENV']['PATH']    = '/usr/local/bin:/opt/bin:/bin:/usr/bin'",
+      "env['ENV']['PATH']    = '#{ENV['HOMEBREW_PREFIX']+'/bin'}:/usr/local/bin:/opt/bin:/bin:/usr/bin'"
```
sjackman: Try using the `/` operator to append `bin` to `HOMEBREW_PREFIX`.
```ruby
"env['ENV']['PATH']    = '#{HOMEBREW_PREFIX/"bin"}:/usr/local/bin:/opt/bin:/bin:/usr/bin'"
```

eiennohito:
Changed the patch as per comments

sjackman:
Merged. Thanks for your contribution, Arseny!

PR #873, opened 2016-03-02
libfuse 2.9.5

dholm:


dholm:
Should I still submit this to Homebrew even though the formula shouldn't be used on OSX?

Library/Formula/libfuse.rb
```diff
@@ -0,0 +1,25 @@
+class Libfuse < Formula
+  desc "Reference implementation of the Linux FUSE (Filesystem in Userspace) interface."
+  homepage "https://github.com/libfuse/libfuse"
+  url "https://github.com/libfuse/libfuse/archive/fuse_2_9_5.tar.gz"
+  sha256 "ccea9c00f7572385e9064bc55b2bfefd8d34de487ba16d9eb09672202b5440ec"
+  head "https://github.com/libfuse/libfuse.git"
+
+  def install
+    system "./makeconf.sh"
+    system "./configure", "--prefix=#{prefix}", "--disable-util"
```
sjackman: Why `--disable-util`?

Library/Formula/libfuse.rb
```diff
@@ -0,0 +1,25 @@
...
+  test do
+    (testpath/"fuse-test.c").write <<-EOS.undent
+      #include <fuse.h>
+      #include <stdlib.h>
+      int main(int argc, char** argv)
+      {
+        struct fuse_operations ops = {NULL, };
+        return fuse_main(argc, argv, &ops, NULL);
+      }
+    EOS
```
sjackman: Did you intend to compile this test as well?

sjackman:
> Should I still submit this to Homebrew even though the formula shouldn't be used on OSX?

No. Mac OS has the formula `osxfuse`.

Library/Formula/libfuse.rb
```diff
@@ -0,0 +1,25 @@
+class Libfuse < Formula
+  desc "Reference implementation of the Linux FUSE (Filesystem in Userspace) interface."
+  homepage "https://github.com/libfuse/libfuse"
+  url "https://github.com/libfuse/libfuse/archive/fuse_2_9_5.tar.gz"
+  sha256 "ccea9c00f7572385e9064bc55b2bfefd8d34de487ba16d9eb09672202b5440ec"
+  head "https://github.com/libfuse/libfuse.git"
```
sjackman: Please add
```ruby
# tag "linuxbrew"
```
I use this comment for formula that are present in Linuxbrew but not in Homebrew.

sjackman:
```
==> audit problems
libfuse:
 * Description is too long. "name: desc" should be less than 80 characters.
Length is calculated as libfuse + desc. (currently 88)
```

> Does your submission pass brew audit --strict <formula> (where <formula> is the name of the formula you're submitting)?

Please read and follow https://github.com/Linuxbrew/linuxbrew/blob/master/.github/PULL_REQUEST_TEMPLATE.md#all-submissions

Library/Formula/libfuse.rb
dholm: It is hardcoded to install into /sbin, for instance /sbin/mount.fuse.  I can patch it to install them to `#{sbin}`.

Library/Formula/libfuse.rb
dholm: Ah, will add!

dholm:
@sjackman Sorry, forgot to audit as well.  Fixed now.

sjackman:
The test fails when using a system compiler (not Homebrewed installed `gcc`). You'll have to add `-I#{include}` and `-L#{lib}` to the compiler command.
```
0.00s==> brew test --verbose libfuse
Testing libfuse
==> /usr/bin/gcc -DFUSE_USE_VERSION=26 -D_FILE_OFFSET_BITS=64 -o fuse-test fuse-test.c -lfuse
fuse-test.c:1:18: fatal error: fuse.h: No such file or directory
compilation terminated.
Error: libfuse: failed
```

sjackman:
Or even better, add `$CPPFLAGS` and `$LDFLAGS` to the compiler command.

dholm:
@sjackman Fixed!

Library/Formula/libfuse.rb
```diff
@@ -0,0 +1,33 @@
...
+      #include <fuse.h>
+      #include <stdlib.h>
+      int main(int argc, char** argv)
+      {
+        struct fuse_operations ops = {NULL, };
+        return fuse_main(argc, argv, &ops, NULL);
+      }
+    EOS
+    cpp_flags = [ENV["CPPFLAGS"], "-DFUSE_USE_VERSION=26",
+                 "-D_FILE_OFFSET_BITS=64"]
```
sjackman: What's the purpose of these two options, `-DFUSE_USE_VERSION=26` and `-D_FILE_OFFSET_BITS=64`?
Can they be omitted?

Library/Formula/libfuse.rb
```diff
@@ -0,0 +1,33 @@
...
+    ld_flags = [ENV["LDFLAGS"], "-lfuse"]
+    system ENV["CC"], *cpp_flags, "-o", "fuse-test", "fuse-test.c", *ld_flags
```
sjackman: Unfortunately `*` doesn't work in the middle of a call, only at the end, with older versions of Ruby. Instead of using temporary variables, expand them inline.

dholm: That explains why I had a vague recollection that it wasn't supposed to work.  :)
The only reason I put them into variables was to keep within a line width of 80 characters, I will inline them.

Library/Formula/libfuse.rb
dholm: From `fuse.h`:

     * IMPORTANT: you should define FUSE_USE_VERSION before including this
     * header.  To use the newest API define it to 26 (recommended for any
     * new application), to use the old API define it to 21 (default) 22
     * or 25, to use the even older 1.X API define it to 11.

As for `_FILE_OFFSET_BITS` this might be an issue that exists in my environment, I will remove it.

Library/Formula/libfuse.rb
sjackman: You can add line breaks after the commas.

sjackman:
The Travis error is transient.
```
Installing rake 10.5.0
Gem::RemoteFetcher::FetchError: Errno::ECONNRESET: Connection reset by peer - SSL_connect (https://rubygems.org/gems/json-1.8.3.gem)
```
https://travis-ci.org/Linuxbrew/linuxbrew/builds/114456010#L228-L230

sjackman:
`libfuse 2.9.5 (new formula) for Linxubrew`

sjackman:
https://gist.github.com/sjackman/00693e477d388d80738a#file-01-makeconf-sh-L14
```
config.rpath not found! - is gettext installed?
```

sjackman:
Added
```ruby
  depends_on "gettext"
  def install
    cp Formula["gettext"].pkgshare/"config.rpath", "."
```

sjackman:
```
/gsc/btl/linuxbrew/include/fuse/fuse_common.h:33:2: error: #error Please add -D_FILE_OFFSET_BITS=64 to your compile flags!
Error: libfuse: failed
```
I guess it was necessary.

sjackman:
The current test fails if `$CPPFLAGS` or `$LDFLAGS` has spaces in it.

sjackman:
This works:
```ruby
    system "$CC $CPPFLAGS -DFUSE_USE_VERSION=26 -D_FILE_OFFSET_BITS=64 fuse-test.c $LDFLAGS -lfuse"
```

sjackman:
Merged. Thanks again for your contribution, David!

PR #879, opened 2016-03-03
cdparanoia: Fix Linux compilation

DoomHammer:
Here are the `brew gist-logs`: https://gist.github.com/anonymous/e766c8c18c05de90644e

### DIFF
--- a/Library/Formula/cdparanoia.rb
+++ b/Library/Formula/cdparanoia.rb
@@ -15,12 +15,17 @@ class Cdparanoia < Formula
   patch do
     url "https://raw.githubusercontent.com/Homebrew/patches/2a22152/cdparanoia/osx_interface.patch"
     sha256 "3eca8ff34d2617c460056f97457b5ac62db1983517525e5c73886a2dea9f06d9"
-  end
+  end if OS.mac?
 
   patch do
     url "https://raw.githubusercontent.com/Homebrew/patches/2a22152/cdparanoia/patch-paranoia_paranoia.c.10.4.diff"
     sha256 "4d6d51909d1b29a3c8ac349f5132a8acd96628355117efb3f192408d2cc4829e"
   end
+
+  patch do
+    url "https://raw.githubusercontent.com/drewc/guix/master/gnu/packages/patches/cdparanoia-fpic.patch"
+    sha256 "496f53d21dde7e23f4c9cf1cc28219efcbb5464fe2abbd5a073635279281c9c4"
+  end if OS.linux?
 
   def install
     system "autoconf"


Library/Formula/cdparanoia.rb
```diff
@@ -15,12 +15,17 @@ class Cdparanoia < Formula
   patch do
     url "https://raw.githubusercontent.com/Homebrew/patches/2a22152/cdparanoia/osx_interface.patch"
     sha256 "3eca8ff34d2617c460056f97457b5ac62db1983517525e5c73886a2dea9f06d9"
-  end
+  end unless OS.linux?
```
sjackman: Please change this to `if OS.mac?`

Library/Formula/cdparanoia.rb
```diff
@@ -15,12 +15,17 @@ class Cdparanoia < Formula
...
 
   patch do
     url "https://raw.githubusercontent.com/Homebrew/patches/2a22152/cdparanoia/patch-paranoia_paranoia.c.10.4.diff"
     sha256 "4d6d51909d1b29a3c8ac349f5132a8acd96628355117efb3f192408d2cc4829e"
-  end
+  end unless OS.linux?
```
sjackman: Ditto.

Library/Formula/cdparanoia.rb
```diff
@@ -15,12 +15,17 @@ class Cdparanoia < Formula
...
+
+  patch do
+    url "https://raw.githubusercontent.com/drewc/guix/master/gnu/packages/patches/cdparanoia-fpic.patch"
+    sha256 "496f53d21dde7e23f4c9cf1cc28219efcbb5464fe2abbd5a073635279281c9c4"
+  end if OS.linux?
```
sjackman: Ditto.

sjackman:
```
/usr/bin/gcc-4.8 -O2 -fsigned-char -Os -w -pipe -march=core2 -c cooked_interface.c
common_interface.c:16:27: fatal error: sys/syslimits.h: No such file or directory
 #include <sys/syslimits.h>
```
https://gist.github.com/anonymous/e766c8c18c05de90644e#file-03-make-L19-L21

sjackman:
Is disabling all three patches necessary? Please only disable those patches necessary to get it to build on Linux.

DoomHammer:
Disabled two, added one. Actually one should be disabled, one left intact and one added. Thanks!

sjackman:
Has the Linux patch been sent to the upstream authors?

sjackman:
```
==> Installing cdparanoia dependency: homebrew/dupes/m4
==> Downloading http://ftpmirror.gnu.org/m4/m4-1.4.17.tar.xz
Already downloaded: /home/travis/.cache/Homebrew/m4-1.4.17.tar.xz
==> Verifying m4-1.4.17.tar.xz checksum
Error: Empty archive
==> FAILED
```
https://travis-ci.org/Linuxbrew/linuxbrew/builds/113594534

sjackman:
Merged. Thanks for your contribution, Piotr.

DoomHammer:
I found this patch first in Debian package, so I believe it has been sent to upstream already.

PR #880, opened 2016-03-03
cdrdao: Fix Linux compilation

DoomHammer:
Gist logs: https://gist.github.com/5c40a554c35054c7178c

### DIFF
--- a/Library/Formula/cdrdao.rb
+++ b/Library/Formula/cdrdao.rb
@@ -20,7 +20,12 @@ class Cdrdao < Formula
   patch do
     url "https://sourceforge.net/p/cdrdao/patches/_discuss/thread/205354b0/141e/attachment/cdrdao-mac.patch"
     sha256 "ee1702dfd9156ebb69f5d84dcab04197e11433dd823e80923fd497812041179e"
-  end
+  end if OS.mac?
+
+  patch do
+    url "https://raw.githubusercontent.com/PacBSD/abs/master/extra/cdrdao/cdrdao-1.2.3-stat.patch"
+    sha256 "ca89b7c56a376d5a9574c5757f0d372236a895334f81867ff5e1703806565bbc"
+  end if OS.linux?
 
   # second patch fixes device autodetection on OS X
   # see https://trac.macports.org/ticket/27819


sjackman:
https://gist.github.com/anonymous/5c40a554c35054c7178c#file-02-make-L164-L167
```
/usr/bin/g++-4.8 -DHAVE_CONFIG_H -I. -I. -I.. -I./../trackdb -I./../paranoia   -isystem/home/doomhammer/.linuxbrew/include -DDRIVER_TABLE_FILE=\"/home/doomhammer/.linuxbrew/Cellar/cdrdao/1.2.3/share/cdrdao/drivers\" -Os -w -pipe -march=core2 -c -o sg_err.o sg_err.cc
ScsiIf-linux.cc: In static member function ‘static ScsiIf::ScanData* ScsiIf::scan(int*, char*)’:
ScsiIf-linux.cc:287:37: error: no matching function for call to ‘stat::stat(const char [22], stat*)’
     if (stat(SYSFS_SCSI_DEVICES, &st) != 0) {
```

Library/Formula/cdrdao.rb
```diff
@@ -20,7 +20,12 @@ class Cdrdao < Formula
   patch do
     url "https://sourceforge.net/p/cdrdao/patches/_discuss/thread/205354b0/141e/attachment/cdrdao-mac.patch"
     sha256 "ee1702dfd9156ebb69f5d84dcab04197e11433dd823e80923fd497812041179e"
-  end
+  end unless OS.linux?
```
sjackman: Please change to `if OS.mac?`

Library/Formula/cdrdao.rb
```diff
@@ -20,7 +20,12 @@ class Cdrdao < Formula
...
+
+  patch do
+    url "https://raw.githubusercontent.com/PacBSD/abs/master/extra/cdrdao/cdrdao-1.2.3-stat.patch"
+    sha256 "ca89b7c56a376d5a9574c5757f0d372236a895334f81867ff5e1703806565bbc"
+  end if OS.linux?
```
sjackman: Ditto.

sjackman:
Is disabling both patches necessary?

DoomHammer:
Um, I only disabled one mac-specific patch. The other one I have added. It is necessary to compile under Linux :)

sjackman:
Ah, yes. Sorry. So many pull requests. Which is a good thing!

sjackman:
Has the Linux patch been submitted to the upstream author?

sjackman:
For future contributions, the preferred commit message summary (the first line) is:
`cdrdao: Fix for Linuxbrew`

sjackman:
Merged. Thanks for your contribution, Piotr!

DoomHammer:
The patch is from a BSD distribution so I believe it has already been posted upstream.

PR #884, opened 2016-03-04
curl: fix for gssapi dependency on linux [Standalone linuxbrew] 

maxim-belkin:
I think the provided fix is necessary for the Standalone installation.
Related [webpage](https://curl.haxx.se/mail/lib-2014-11/0187.html)

The command being executed: `brew install curl --with-gssapi`

Without the fix:
gist logs [here](https://gist.github.com/36a3011b62f48e16ff35)
```bash
$ ldd [Linuxbrew]/Cellar/curl/7.47.1/bin/curl
linux-vdso.so.1 (0x00007fffc8dff000)
libcurl.so.4 => [Linuxbrew]/lib/libcurl.so.4 (0x00007fa0f99db000)
libssl.so.1.0.0 => [Linuxbrew]/lib/libssl.so.1.0.0 (0x00007fa0f976a000)
libcrypto.so.1.0.0 => [Linuxbrew]/lib/libcrypto.so.1.0.0 (0x00007fa0f9317000)
libz.so.1 => [Linuxbrew]/lib/libz.so.1 (0x00007fa0f9105000)
libc.so.6 => [Linuxbrew]/lib/libc.so.6 (0x00007fa0f8d8c000)
libdl.so.2 => [Linuxbrew]/lib/libdl.so.2 (0x00007fa0f8b88000)
[Linuxbrew]/lib/ld.so (0x00007fa0f9c2f000)
```

With the fix:
gist logs [here](https://gist.github.com/8448af14bb610ca1d574)
```bash
$ ldd [Linuxbrew]/Cellar/curl/7.47.1/bin/curl
linux-vdso.so.1 (0x00007fff43579000)
libcurl.so.4 => [Linuxbrew]/lib/libcurl.so.4 (0x00007ffa258f8000)
libssl.so.1.0.0 => [Linuxbrew]/lib/libssl.so.1.0.0 (0x00007ffa25687000)
libcrypto.so.1.0.0 => [Linuxbrew]/lib/libcrypto.so.1.0.0 (0x00007ffa25234000)
libz.so.1 => [Linuxbrew]/lib/libz.so.1 (0x00007ffa25022000)
libc.so.6 => [Linuxbrew]/lib/libc.so.6 (0x00007ffa24ca9000)
libgssapi_krb5.so.2 => [Linuxbrew]/lib/libgssapi_krb5.so.2 (0x00007ffa24a6b000)
libkrb5.so.3 => [Linuxbrew]/lib/libkrb5.so.3 (0x00007ffa247b0000)
libk5crypto.so.3 => [Linuxbrew]/lib/libk5crypto.so.3 (0x00007ffa24585000)
libcom_err.so.3 => [Linuxbrew]/lib/libcom_err.so.3 (0x00007ffa24382000)
libdl.so.2 => [Linuxbrew]/lib/libdl.so.2 (0x00007ffa2417e000)
[Linuxbrew]/lib/ld.so (0x00007ffa25b50000)
libkrb5support.so.0 => [Linuxbrew]/lib/libkrb5support.so.0 (0x00007ffa23f74000)
libresolv.so.2 => [Linuxbrew]/lib/libresolv.so.2 (0x00007ffa23d5f000
```

**Update**:
Look at the end of config (search for "curl version:     7.47.1") and see `GSS-API support`

**Update 2**:
I was able to successfully build curl with GSSAPI support without the change introduced on line 85.
If someone desires to use system's gssapi, then change on line 85 has to go. But I lean towards keeping that change.

It would be nice to have `:standalone_linuxbrew` macro similar to `:build` and others...

### DIFF
--- a/Library/Formula/curl.rb
+++ b/Library/Formula/curl.rb
@@ -37,6 +37,7 @@ class Curl < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "homebrew/dupes/krb5" => :build if build.with?("gssapi") unless OS.mac?
   depends_on "libidn" => :optional
   depends_on "rtmpdump" => :optional
   depends_on "libssh2" => :optional
@@ -81,7 +82,7 @@ class Curl < Formula
     args << (build.with?("libssh2") ? "--with-libssh2" : "--without-libssh2")
     args << (build.with?("libidn") ? "--with-libidn" : "--without-libidn")
     args << (build.with?("libmetalink") ? "--with-libmetalink" : "--without-libmetalink")
-    args << (build.with?("gssapi") ? "--with-gssapi" : "--without-gssapi")
+    args << (build.with?("gssapi") ? "--with-gssapi" + (OS.mac? ? "" : "=#{Formula["homebrew/dupes/krb5"].opt_prefix}") : "--without-gssapi")
     args << (build.with?("rtmpdump") ? "--with-librtmp" : "--without-librtmp")
 
     if build.with? "c-ares"


Library/Formula/curl.rb
```diff
@@ -37,6 +37,7 @@ class Curl < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "homebrew/dupes/krb5" => :build if build.with?("gssapi") unless OS.mac?
```
sjackman: Are you sure this is a `:build` dependency and not also a run-time dependency?

Library/Formula/curl.rb
```diff
@@ -81,7 +82,7 @@ def install
     args << (build.with?("libssh2") ? "--with-libssh2" : "--without-libssh2")
     args << (build.with?("libidn") ? "--with-libidn" : "--without-libidn")
     args << (build.with?("libmetalink") ? "--with-libmetalink" : "--without-libmetalink")
-    args << (build.with?("gssapi") ? "--with-gssapi" : "--without-gssapi")
+    args << (build.with?("gssapi") ? "--with-gssapi" + (OS.mac? ? "" : "=#{Formula["homebrew/dupes/krb5"].opt_prefix}") : "--without-gssapi")
```
sjackman: Is this line necessary? Please post a gist of how it fails without this line.

sjackman:
I may make `--without-gssapi` the default on Linuxbrew. Thoughts?

Library/Formula/curl.rb
sjackman: > Update 2:
I was able to successfully build curl with GSSAPI support without the change introduced on line 85.
If someone desires to use system's gssapi, then change on line 85 has to go. But I lean towards keeping that change.

Library/Formula/curl.rb
sjackman: It is a run-time dependency.
```
libkrb5.so.3 => [Linuxbrew]/lib/libkrb5.so.3 (0x00007ffa247b0000)
```

sjackman:
What is GSSAPI support used for? What do you use it for? I'm just curious. I've never used it.

Library/Formula/curl.rb
maxim-belkin: Yes, once dependency on `krb5` is added change on line 85 is not necessary. My concern is that `curl` still builds even if it does not find `gssapi` when `--with-gssapi` is provided.

maxim-belkin:
It is needed to make curl understand [Kerberos](http://www.cloudera.com/documentation/archive/cdh/4-x/4-2-0/CDH4-Security-Guide/cdh4sg_topic_7_2.html). I'm fine with it being turned off by default.

Library/Formula/curl.rb
maxim-belkin: Here is an exert from `./configure --help` for curl:

```text
  --with-gssapi-includes=DIR
                          Specify location of GSS-API headers
  --with-gssapi-libs=DIR  Specify location of GSS-API libs
  --with-gssapi=DIR       Where to look for GSS-API
```

sjackman:
What do you personally use Kerberos for though? I've never found an occasion to use it.

sjackman:
Merged. Thanks again for your contribution, Maxim.

PR #885, opened 2016-03-04
kdiff3: fix for Linuxbrew

rwhogg:
Message:
> Error: No such file or directory - releaseQt/kdiff3.app


sjackman:
For future contributions the preferred commit message is `kdiff3: Fix for Linuxbrew`

sjackman:
Merged. Thanks, Bob!

PR #886, opened 2016-03-04
Update python.rb

erigler-usgs:
I added `-L#{HOMEBREW_PREFIX}/lib` to the LDFLAGS added to the argument list for configure. This is so that Python's setup.py can find "necessary bits" to build extensions. According to a note in Python setup.py's self.detect_modules(), LDFLAGS (along with CPPFLAGS) is somehow pulled from the Makefile, which means that these values must be passed as arguments to configure, not just set in the environment. This seems to explain why including these in std.rb did not work as expected.


sjackman:
Merged. Thanks for your contribution!

PR #887, opened 2016-03-04
gcc: remove 'lib.mkdir' for gcc-4.4

maxim-belkin:
Fix for `gcc` when built with glibc
```sh
Errno::EEXIST: File exists - [Linuxbrew]/Cellar/gcc/5.3.0/lib
```


sjackman:
Merged `mkdir_p lib`. Thanks!

PR #888, opened 2016-03-04
libmetalink: add libxml2 build dependency on Linux

maxim-belkin:
`libmetalink` requires either `expat` or `libxml2` to successfully build on Linux. Here I add `libxml2` as a build dependency on Linux.

### DIFF
--- a/Library/Formula/libmetalink.rb
+++ b/Library/Formula/libmetalink.rb
@@ -12,6 +12,7 @@ class Libmetalink < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "libxml2" => :build unless OS.mac?
 
   def install
     system "./configure", "--disable-dependency-tracking",


sjackman:
For future contributions the preferred commit message is `libmetalink: Fix for Linuxbrew`

sjackman:
Please a log of the failed build without your patch using `brew gist-logs libmetalink`.
See https://github.com/Linuxbrew/linuxbrew/blob/master/.github/PULL_REQUEST_TEMPLATE.md

sjackman:
```
checking for EXPAT... no
configure: WARNING: No package 'expat' found
checking for XML_ParserCreate in -lexpat... no
checking for xml2-config... /usr/bin/xml2-config
checking for libxml - version >= 2.6.24... no
*** Could not run libxml test program, checking why...
*** The test program failed to compile or link. See the file config.log for the
*** exact error that occured. This usually means LIBXML was incorrectly installed
*** or that you have moved LIBXML since it was installed. In the latter case, you
*** may want to edit the xml2-config script: /usr/bin/xml2-config
configure: error: in `/var/tmp/sjackman/libmetalink20160304-1573-qizhe1/libmetalink-0.1.3':
configure: error: Neither expat nor libxml2 found. Please install one of them and run configure again.
```

sjackman:
`libmetalink: Depends on libxml2 for Linuxbrew`

Library/Formula/libmetalink.rb
```diff
@@ -12,6 +12,7 @@ class Libmetalink < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "libxml2" => :build unless OS.mac?
```
sjackman: Is it definitely a `:build` dependency and not also a run-time dependency?

sjackman:
Do you have a particular reason to choose `libxml2` over `expat`?

Library/Formula/libmetalink.rb
sjackman: It is a run-time dependency also.
```
$ ldd /home/sjackman/.linuxbrew/Cellar/libmetalink/0.1.3/lib/libmetalink.so
	linux-vdso.so.1 (0x00007fffc8b5c000)
	libxml2.so.2 => /home/sjackman/.linuxbrew/lib/libxml2.so.2 (0x00002ba89cc04000)
	libz.so.1 => /home/sjackman/.linuxbrew/lib/libz.so.1 (0x00002ba89cf2a000)
	libm.so.6 => /home/sjackman/.linuxbrew/lib/libm.so.6 (0x00002ba89d13d000)
	libdl.so.2 => /home/sjackman/.linuxbrew/lib/libdl.so.2 (0x00002ba89d430000)
	libc.so.6 => /home/sjackman/.linuxbrew/lib/libc.so.6 (0x00002ba89d634000)
	/home/sjackman/.linuxbrew/Cellar/glibc/2.19/lib64/ld-linux-x86-64.so.2 (0x00002ba89c7d5000)
```

sjackman:
Lucky issue #888!

sjackman:
Merged. Thanks again for your contribution, Maxim! Have a good weekend.

maxim-belkin:
Glad I can help. Chose of `libxml2` over `expat` is random. There should be an option to build either with `expat` or with `libxml2`, or with both :)

PR #890, opened 2016-03-04
capstone: Fix Linux install

dholm:


dholm:
I was told to resubmit here?
https://github.com/Homebrew/homebrew/pull/49779

sjackman:
For future contributions the preferred commit message is `capstone: Fix for Linuxbrew`

sjackman:
Merged as
```ruby
    system "./make.sh"
    if OS.mac?
      system "./make.sh", "install"
    else
      system "make", "install", "PREFIX=#{prefix}"
    end
```

sjackman:
Thanks again for your contribution, David!

PR #891, opened 2016-03-05
WORK IN PROGRESS libxcb 1.11.1 (new formula)

rwhogg:
This adds `libxcb`, a new formula for the client side of X11. It also includes the direct dependencies `libxau` and `xproto`.

TODO:
* [ ] Split resources out into separate formulae
* [X] Write tests
* [X] Verify all dependencies are 100% correct
* [X] Tag the formulae as exclusive to Linuxbrew
* [X] Update the versions - some of them are outdated

Issue #826, opened 2016-02-19
tcl-tk: Tk depends on X11


maxim-belkin:
I've been trying to install `tk` part of the `tcl-tk` (installation of `tcl` part proceeds smoothly).
Things I've found so far:
1. `with-x11` option uses incorrect syntax `--with-x` for `configure` command: it should be `--with-x=DIR`.
2. `patch: DATA` checks `if MacOS.version < :leopard` which should be prepended with `OS.mac? &&`
3. To find `Xlib.h` (actually, we need `X11/Xlib.h`), I used:
`rpm -ql libX11-devel | grep Xlib.h`
(system is CentOS 6.6), which returned
`/usr/include/X11/Xlib.h`
So, we need the compiler to know about `/usr/include` folder. If we append `/usr/include` to the `CFLAGS` using `ENV.append_to_cflags "-I/usr/include"` then the compiler finds an old `tcl.h` and installation fails. So, I create a link to that `X11` folder using
`system "ln", "-s", "/usr/include/X11", Dir.pwd`
as the first step of the `tk`-building phase (before the `configure` command).
4. At this point the build process fails with an error about missing `libX11.so`. In my case it is in `/usr/lib64`. So, I append this path to cflags with `ENV.append_to_cflags "-L/usr/lib64"`.
5. At this point the build process fails with an error message about unknown `libxcb.so`....
And this is where things seem weird:
```
$ ldd /usr/lib64/libX11.so
	linux-vdso.so.1 (0x00007fffd37ff000)
	libxcb.so.1 => not found
	libdl.so.2 => [linuxbrew-folder]/Cellar/glibc/2.19/lib/libdl.so.2 (0x00007f14b0c24000)
	libc.so.6 => [linuxbrew-folder]/Cellar/glibc/2.19/lib/libc.so.6 (0x00007f14b08aa000)
	[linuxbrew-folder]/Cellar/glibc/2.19/lib64/ld-linux-x86-64.so.2 (0x00007f14b0e2c000)

$ /usr/bin/ldd /usr/lib64/libX11.so
	linux-vdso.so.1 =>  (0x00007fffb9fbd000)
	libxcb.so.1 => /usr/lib64/libxcb.so.1 (0x0000003625400000)
	libdl.so.2 => /lib64/libdl.so.2 (0x0000003622c00000)
	libc.so.6 => /lib64/libc.so.6 (0x0000003622000000)
[...]

PR #892, opened 2016-03-06
Xorg libraries: new formula [WIP]

maxim-belkin:
Inspired by the pull request #891 and remembering my promise back in #826, I've finally decided to spent time to create `Xorg libraries` formula according to [LFS](http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html). I've tested it on my machine, it works. The problem is that formula has to have a common URL that is used to deduce version number and Xorg libraries don't have a single url. I copied info from one of the Xorg "resources" (libFS I believe), so xorg is installed in a subfolder `1.0.7` instead of just `7`. So, this part needs fix and I'm not sure how to do that.

### DIFF
--- a/Library/Formula/xorg.rb
+++ b/Library/Formula/xorg.rb
@@ -1,6 +1,6 @@
 class Xorg < Formula
   desc "Xorg Libraries"
-  homepage "http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html"
+  homepage "http://www.x.org/" ### http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html
   version "7"
 
   url "http://ftp.x.org/pub/individual/lib/libFS-1.0.7.tar.bz2"
@@ -17,12 +17,25 @@ class Xorg < Formula
 # depends_on "libxcb"      # built as a resource here
 
 # Xorg: optional dependencies
-  if build.with?("docs") 
-    depends_on "xmlto"
-    depends_on "fop"        =>  :optional # needs update to 2.1
-    depends_on "Links"      =>  :optional
-    depends_on "lynx"       =>  :optional
-    depends_on "w3m"        =>  :optional
+  if build.with?("docs")
+    depends_on "xmlto"      =>  :build
+
+    option "with-fop",   "Use fop to build documentation"
+    option "with-Links", "Use Links to build documentation"
+    option "with-lynx",  "Use lynx to build documentation"
+    option "with-w3m",   "Use w3m to build documentation"
+
+    if !build.with?("fop")  and !build.with?("Links") and !build.with?("lynx")  and !build.with?("w3m")
+      depends_on "fop"        =>  :optional
+      depends_on "Links"      =>  :optional
+      depends_on "lynx"       =>  :optional
+      depends_on "w3m"        =>  :optional
+    end
+
+    depends_on "fop"        =>  :build if build.with?("fop") # needs update to 2.1
+    depends_on "Links"      =>  :build if build.with?("Links")
+    depends_on "lynx"       =>  :build if build.with?("lynx")
+    depends_on "w3m"        =>  :build if build.with?("w3m")
   end
 
 # libxcb: required dependencies
@@ -30,14 +43,14 @@ class Xorg < Formula
 # depends_on "xcbproto"    # built as a resource here
 
 # libxcb: optional dependencies
-  depends_on "doxygen"    =>  :optional if build.with?("doxygen") # to build API docs
-  depends_on "check"      =>  :optional if build.with?("check-libxcb") # to run tests
+  depends_on "doxygen"    =>  :build if build.with?("doxygen") # to build API docs
+  depends_on "check"      =>  :build if build.with?("check-libxcb") # to run tests
   depends_on "libxslt"    =>  :optional # no explanation given
 
 
 # xcbproto: required dependencies
   depends_on :python          # 
-  depends_on "libxml2"    =>  :optional if build.with?("check-xcbproto") # to run tests
+  depends_on "libxml2"    =>  :build if build.with?("check-xcbproto") # to run tests
  #depends_on "libXdmcp"   => [:optional, :recommended] # built as a resource here
   depends_on "pkg-config" => :build
 
@@ -430,7 +443,10 @@ class Xorg < Formula
     end
 
 ### libpthread-stubs
-### see inreplace below 
+###
+### see inreplace below ###
+### http://www.linuxfromscratch.org/blfs/view/svn/x/libxcb.html
+###
 ##    resource("libpthread-stubs").stage do
 ##      system "./configure", "--prefix=#{prefix}"
 ##      system "make"
@@ -464,7 +480,7 @@ class Xorg < Formula
 
         system "make"
         system "make", "install"
-        system "ldconfig"
+        # system "ldconfig"  ### https://github.com/Linuxbrew/linuxbrew/pull/892#discussion_r55162757
       end
     end
   end
--- a/Library/Formula/xorg.rb
+++ b/Library/Formula/xorg.rb
@@ -1,0 +1,472 @@
+class Xorg < Formula
+  desc "Xorg Libraries"
+  homepage "http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html"
+  version "7"
+
+  url "http://ftp.x.org/pub/individual/lib/libFS-1.0.7.tar.bz2"
+  sha256 "2e9d4c07026a7401d4fa4ffae86e6ac7fec83f50f3268fa85f52718e479dc4f3"
+  # tag "linuxbrew"
+
+  option "with-docs", "Build additional PDF or text documentation for the libXfont"
+  option "with-doxygen", "Generate API documentation (for libxcb)"
+  option "with-check-libxcb", "Run tests (for libxcb)"
+  option "with-check-xcbproto", "Run tests (for xcb-proto)"
+
+# Xorg: required dependencies
+  depends_on "fontconfig"
+# depends_on "libxcb"      # built as a resource here
+
+# Xorg: optional dependencies
+  if build.with?("docs") 
+    depends_on "xmlto"
+    depends_on "fop"        =>  :optional # needs update to 2.1
+    depends_on "Links"      =>  :optional
+    depends_on "lynx"       =>  :optional
+    depends_on "w3m"        =>  :optional
+  end
+
+# libxcb: required dependencies
+# depends_on "libXau"      # built as a resource here
+# depends_on "xcbproto"    # built as a resource here
+
+# libxcb: optional dependencies
+  depends_on "doxygen"    =>  :optional if build.with?("doxygen") # to build API docs
+  depends_on "check"      =>  :optional if build.with?("check-libxcb") # to run tests
+  depends_on "libxslt"    =>  :optional # no explanation given
+
+
+# xcbproto: required dependencies
+  depends_on :python          # 
+  depends_on "libxml2"    =>  :optional if build.with?("check-xcbproto") # to run tests
+ #depends_on "libXdmcp"   => [:optional, :recommended] # built as a resource here
+  depends_on "pkg-config" => :build
+
+  depends_on "gettext"    => :build
+##################################
+
+  ### PROTO
+  
+  resource "util-macros" do
+    url "http://www.x.org/archive/individual/util/util-macros-1.19.0.tar.bz2"
+    sha256 "2835b11829ee634e19fa56517b4cfc52ef39acea0cd82e15f68096e27cbed0ba"
+  end
+
+  ##### XORG PROTOCOL HEADERS
+
+  resource "bigreqsproto" do
+    url      "http://ftp.x.org/pub/individual/proto/bigreqsproto-1.1.2.tar.bz2"
+    sha256   "462116ab44e41d8121bfde947321950370b285a5316612b8fce8334d50751b1e"
+  end
+
+  resource "compositeproto" do
+    url      "http://ftp.x.org/pub/individual/proto/compositeproto-0.4.2.tar.bz2"
+    sha256   "049359f0be0b2b984a8149c966dd04e8c58e6eade2a4a309cf1126635ccd0cfc"
+  end
+
+  resource "damageproto" do
+    url      "http://ftp.x.org/pub/individual/proto/damageproto-1.2.1.tar.bz2"
+    sha256   "5c7c112e9b9ea8a9d5b019e5f17d481ae20f766cb7a4648360e7c1b46fc9fc5b"
+  end
+
+  resource "dmxproto" do
+    url      "http://ftp.x.org/pub/individual/proto/dmxproto-2.3.1.tar.bz2"
+    sha256   "e72051e6a3e06b236d19eed56368117b745ca1e1a27bdc50fd51aa375bea6509"
+  end
+
+  resource "dri2proto" do
+    url      "http://ftp.x.org/pub/individual/proto/dri2proto-2.8.tar.bz2"
+    sha256   "f9b55476def44fc7c459b2537d17dbc731e36ed5d416af7ca0b1e2e676f8aa04"
+  end
+
+  resource "dri3proto" do
+    url      "http://ftp.x.org/pub/individual/proto/dri3proto-1.0.tar.bz2"
+    sha256   "01be49d70200518b9a6b297131f6cc71f4ea2de17436896af153226a774fc074"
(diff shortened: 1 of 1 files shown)


Library/Formula/xorg.rb
```diff
@@ -0,0 +1,470 @@
...
+  end
+
+  ##### END of XPROTOCOL HEADERS 
+
+  resource "xcb-proto" do
+    url "http://xcb.freedesktop.org/dist/xcb-proto-1.11.tar.gz"
+    sha256 "d12152193bd71aabbdbb97b029717ae6d5d0477ab239614e3d6193cc0385d906"
+  end
+
+## for libxcb... not sure where it is needed there
```
rwhogg: Strictly speaking, libpthread-stubs isn't necessary on Linux. [BLFS](http://www.linuxfromscratch.org/blfs/view/svn/x/libxcb.html) actually recommends stripping the dependency on it out:

> sed "s/pthread-stubs//" -i configure: This sed removes dependency on libpthread-stubs package which is useless on Linux. 

I wasn't wholly comfortable with that, so I decided to just include it as a resource.

rwhogg: Ah, I see that you already accounted for that.

rwhogg:
I'm getting the following error (ran this on a standalone-Linuxbrew Docker image)
> make[2]: Leaving directory '/tmp/xorg--libXpm20160306-30770-8m7m98/libXpm-3.5.11/src'
Making all in sxpm
make[2]: Entering directory '/tmp/xorg--libXpm20160306-30770-8m7m98/libXpm-3.5.11/sxpm'
  CC       sxpm.o
  GEN      sxpm.po
/bin/bash: xgettext: command not found
Makefile:651: recipe for target 'sxpm.po' failed
make[2]: *** [sxpm.po] Error 127
make[2]: Leaving directory '/tmp/xorg--libXpm20160306-30770-8m7m98/libXpm-3.5.11/sxpm'
Makefile:467: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/tmp/xorg--libXpm20160306-30770-8m7m98/libXpm-3.5.11'
Makefile:375: recipe for target 'all' failed
make: *** [all] Error 2

Also, FYI https://github.com/Homebrew/homebrew/pull/49497 the xorg macros might be coming to plain Homebrew and it looks like they're discussing some of the same issues we are here.

Library/Formula/xorg.rb
```diff
@@ -0,0 +1,470 @@
...
+  depends_on "doxygen"    =>  :optional if build.with?("doxygen") # to build API docs
+  depends_on "check"      =>  :optional if build.with?("check-libxcb") # to run tests
+  depends_on "libxslt"    =>  :optional # no explanation given
+
+
+# xcbproto: required dependencies
+  depends_on :python          # 
+  depends_on "libxml2"    =>  :optional if build.with?("check-xcbproto") # to run tests
+ #depends_on "libXdmcp"   => [:optional, :recommended] # built as a resource here
+# depends_on "pkg-config" => :build
```
rwhogg: May I ask why this is commented out?

Library/Formula/xorg.rb
```diff
@@ -0,0 +1,470 @@
+class Xorg < Formula
+  desc "Xorg Libraries"
+  homepage "http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html"
+  revision 7
```
rwhogg: I don't think we need this. Quoting from the [Formula Cookbook](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#formula-cookbook):

> Occasionally, these updates require a forced-recompile of the formula itself or its dependents to either ensure formulae continue to function as expected or to close a security issue. This forced-recompile is known as a revision and inserted underneath the homepage/url/sha block.

Library/Formula/xorg.rb
```diff
@@ -0,0 +1,470 @@
+class Xorg < Formula
+  desc "Xorg Libraries"
+  homepage "http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html"
```
rwhogg: Strictly a matter of opinion, but I think http://www.x.org/wiki/ would make more sense for the homepage.

Library/Formula/xorg.rb
maxim-belkin: Agreed. I was trying to make xorg install into `7` subdirectory.

Library/Formula/xorg.rb
```diff
@@ -0,0 +1,470 @@
...
+  # tag "linuxbrew"
+
+  option "with-docs", "Build additional PDF or text documentation for the libXfont"
+  option "with-doxygen", "Generate API documentation (for libxcb)"
+  option "with-check-libxcb", "Run tests (for libxcb)"
+  option "with-check-xcbproto", "Run tests (for xcb-proto)"
+
+# Xorg: required dependencies
+  depends_on "fontconfig"
+# depends_on "libxcb"      # built as a resource here
```
rwhogg: Not sure exactly where to put it, but I think you're missing a dependency on `gettext`.

Library/Formula/xorg.rb
maxim-belkin: Agree. I used this web page because I followed steps from there.

Library/Formula/xorg.rb
maxim-belkin: could be. I will add it. I'll check where it comes from.

Library/Formula/xorg.rb
maxim-belkin: Well, the main reason is that I was finishing it up at 5am.

rwhogg:
@maxim-belkin I can confirm that, with the added dependency on `gettext`, this installs correctly on a standalone Linuxbrew Docker container.

Great work here!

Library/Formula/xorg.rb
sjackman: Use
```ruby
version "7"
```

sjackman:
Great work! Thanks Maxim and Bob. It's going to be very useful, I think.

What's the thoughts on one big formula with many resources vs one formula per library? I have a uninformed preference toward one formula per library and one `xorg` virtual package that depends on it, but I'd like to hear the benefits of both ways.

maxim-belkin:
I'd prefer one virtual package and many individual ones as small packages are easy to update. But the trouble is that there are 60+ of them (of really small ones). We can probably create all the small ones and split them into meta-packages as in LFS and Xorg will use those meta-packages.

rwhogg:
Given that so many of the individual formulae use the exact same `configure` flags and whatnot and aren't likely to be depended on directly, this single formula idea actually makes a lot of sense to me. (Ironically, me @maxim-belkin are both arguing for the opposite of what our respective PR's actually do!)

There's also the matter of boilerplate - all of these small formulae would probably duplicate details.

Finally, given that upstream Homebrew has PR's like https://github.com/Homebrew/homebrew/pull/49497, splitting up the formulae sounds like it could potentially conflict. By contrast, a single `xorg` formula probably won't do that, since XQuartz seems to be the de facto standard on Macs.

Library/Formula/xorg.rb
```diff
@@ -0,0 +1,470 @@
...
+      system "make", "install"
+    end
+
+ 
+    res = %w[xtrans libX11 libXext libFS libICE libSM libXScrnSaver libXt libXmu libXpm libXaw libXfixes libXcomposite libXrender libXcursor libXdamage libfontenc libXfont libXft libXi libXinerama libXrandr libXres libXtst libXv libXvMC libXxf86dga libXxf86vm libdmx libpciaccess libxkbfile libxshmfence]
+
+    res.each do |r|
+      resource(r).stage do
+
+      case r
```
sjackman: A hash table could be used here rather than a `case` statement.

Library/Formula/xorg.rb
```diff
@@ -0,0 +1,470 @@
...
+        system "./configure", "--prefix=#{prefix}", "--sysconfdir=#{prefix}/etc", "--localstatedir=#{prefix}/var", "--disable-static", "--disable-devel-docs"
+      when "libXt"
+        system "./configure", "--prefix=#{prefix}", "--sysconfdir=#{prefix}/etc", "--localstatedir=#{prefix}/var", "--disable-static", "--with-appdefaultdir=#{prefix}/X11/app-defaults"
+      else
+        system "./configure", "--prefix=#{prefix}", "--sysconfdir=#{prefix}/etc", "--localstatedir=#{prefix}/var", "--disable-static"
+      end
+
+        system "make"
+        system "make", "install"
+        system "ldconfig"
```
sjackman: `ldconfig` shouldn't be necessary. Linuxbrew uses `RPATH` rather than `ld.so.conf`.

sjackman:
The boilerplate code could be deduplicated by using a `XFormula` base class. It would be easier to update the versions of the individual components if they were in separate formula. It could be a `Linuxbrew/x11` tap with the 60+ formula in it.

I am okay though with one big `xorg` package, or a perhaps a handful of packages as Maxim was suggesting. With one big formula, it might be easier to update it periodically on some sort of release schedule, and tie those updates to the release schedule of the XOrg project.

Library/Formula/xorg.rb
maxim-belkin: I think `case` is easier to understand in this case because of `else`.

maxim-belkin:
Interestingly, on OS X the formula uses `install -c -d` to create directories, while on Linux it uses `mkdir -p`. As a result, `xextproto` fails to build because of repeating directory creation somewhere...

maxim-belkin:
Adding `depends_on "coreutils" => :build if OS.mac?` fixes it!

sjackman:
Interesting, and how odd.

sjackman:
And cool to hear that it builds on Mac OS! I was wondering whether it would.

sjackman:
`version "7"`
Can you find out what versions of these libraries are used by `xorg 7.7`, use those same versions of the libraries, and change the version number to `7.7`?

maxim-belkin:
I can certainly do that but... the packages in xorg 7.7 are outdated. The following statement can be found on [Xorg website](http://www.x.org/wiki/Releases/7.7/)

> Update: Many modules have had new releases since the X11R7.7 katamari - see the individual X.Org releases directory for downloads, and the xorg-announce archives for details on included changes. These updates may be included in a future 7.8 katamari release if one is ever assembled.

So, do we want to downgrade the packages? We could create a formula `xorg-7.7` and `xorg-latest`...

There is still a problem with Mac OS, actually. For whatever reason, `binutils` on a Mac do not provide `ld` (as well as `as` and `gprof`). Moreover, `gcc` package on a Mac does not create symlinks `gcc -> gcc-5`, `cpp -> cpp-5` that are created on a Linux. Absence of these cause all sorts of problems later on. I'm sure it is solvable though.

maxim-belkin:
So, it looks like the trouble on OS X is caused by non-GNU ld. I will try to follow [these guides](https://github.com/cfenollosa/os-tutorial/tree/master/11-kernel-crosscompiler) to make it compile.

maxim-belkin:
Question: I've compiled `binutils` on OS X so that they have ld and correct gcc. How do I use that `ld` instead of a system's one? For cc, I can use `--cc=gcc`, but what about `ld`? I've tried setting it inside of a formula (ENV["LD"] = "/path/to/ld") and it didn't work.

sjackman:
Hi, Maxim. On Mac OS it should use the `binutils` provided by the host OS (aka XCode and Command Line Tools). If it can build with `clang`, great, and if not, add a `fails_with :clang` stanza. If it doesn't build with the host `binutils` I would forget about the Mac OS support for now and focus on Linux.

sjackman:
If we're releasing one big X.Org formula, I want to avoid updating it super often and driving the users crazy with updates. If we're going to do only periodic releases, it would be better I think to tie our releases to X.Org releases. [X.Org 7.7](http://www.x.org/wiki/Releases/7.7/) was released on 2012-06-06, so it's pretty darn ancient. So, I agree I'd rather not use X.Org 7.7 What do you think we should do? What has Ubuntu done? How about we use the same versions as a particular Ubuntu release, such as trusty or wily.

sjackman:
> Actually, problems on a Mac appear very close to the very end of compilation and this is where I felt that if we had small packages -- it would have been much faster to troubleshoot.

There's a number of things that may be easier with small packages. Packaging each library individually in `linxubrew/x11` tap is still appealing to me.

maxim-belkin:
`linuxbrew/x11` seems like a good idea to me too. Let's do that. Will you create this tap first? I could then make a PR to it.  I'm not a fan of outdated packages so I vote for having the latest xorg.

maxim-belkin:
There is a problem with Mac's `cpp`, not `clang`. We can indeed work on it after releasing a version for Linux.

sjackman:
Homebrew already has a tap named `homebrew/x11`, which is for application that use X11 libraries.
To avoid confusion, I'm going to start a tap named `linuxbrew/xorg`, for the X.Org libraries themselves.

sjackman:
Done: https://github.com/Linuxbrew/homebrew-xorg

sjackman:
@rwhogg Which do you prefer, the monolithic `xorg` formula or the `Linuxbrew/xorg` tap?

maxim-belkin:
Great, I've tapped it. I will start working on the formulas then.

rwhogg:
I think a tap sounds good.

On March 8, 2016 11:07:33 AM PST, Shaun Jackman <notifications@github.com> wrote:
>@rwhogg Which do you prefer, the monolithic `xorg` formula or the
>`Linuxbrew/xorg` tap?
>
>---
>https://github.com/Linuxbrew/linuxbrew/pull/892#issuecomment-193919900

-- 

sjackman:
Great.

maxim-belkin:
Is there a clean way to define arguments shared across multiple formulas?
In all small formulas we will be using something like: 
```ruby
    args = %W[
      --prefix=#{HOMEBREW_PREFIX}/Cellar/xorg/latest
      --sysconfdir=#{HOMEBREW_PREFIX}/Cellar/xorg/latest/etc
      --localstatedir=#{HOMEBREW_PREFIX}/Cellar/xorg/latest/var
    ]
    system "./configure", *args
    system "make", "install"
```

Or do you think we should install formulas into their own folders (`--prefix=#{prefix}`)? LFS guide uses the same prefix for all the formulas. So, if we can't define common arguments in one place -- we have to make sure everything plays well.

maxim-belkin:
I will use `#{prefix}` for now and then will change that if necessary.

sjackman:
> Or do you think we should install formulas into their own folders (--prefix=#{prefix})?

Definitely put each formula in its own prefix.

maxim-belkin:
Quick question: how to avoid "Error: Empty installation" in a meta-formula that does not have anything to do inside its ```def install```?

maxim-belkin:
I've figured this out. I just need to create a file that is not recognized by Homebrew as unnesessary.

maxim-belkin:
I'm almost done with the packages...

sjackman:
> Quick question: how to avoid "Error: Empty installation" in a meta-formula that does not have anything to do inside its def install?

Add a `README.md` file that describes the purpose of the meta-formula. Put it either in `prefix` or `doc`.

> I'm almost done with the packages...
 Hide all checks

Great!

maxim-belkin:
Looking at `Library/Homebrew/metafiles.rb` it looks like` README.md` will be ignored. So I just create an `html` file in there with different name.

maxim-belkin:
Stylistic question:

With a single formula for `xorg` I used **`--sysconfdir=#{prefix}/etc`** and **`--localstatedir=#{prefix}/var`** which were ending up in 
**`[linuxbrew]/Cellar/xorg/latest/etc`**  and    **`[linuxbrew]/Cellar/xorg/latest/var`**.
With multiple formulas, I can imagine a situation when someone needs just a single formula not the entire set of xorg libraries. Therefore, a question: should I use
    * **`#{etc}`** and **`#{var}`** (i.e. *global* Homebrew folders),   or
    * **`#{prefix}/etc`** and **`#{prefix}/var`** (local to each formula),    or
    *  **`#{HOMEBREW_PREFIX}/Cellar/xorg/latest/etc`** and **`#{HOMEBREW_PREFIX}/Cellar/xorg/latest/var`** (*common* for all formulas)?

sjackman:
I believe #{etc} and #{var} (i.e. global Homebrew folders).

sjackman:
Moved to Linuxbrew/homebrew-xorg/issues/1.

sjackman:
@tseemann Torsten, I recall you having some X issues with R. Have you tried Linuxbrew/xorg/xorg?

PR #900, opened 2016-03-09
foremost: Fix for Linuxbrew

rwhogg:
[x] Have you followed the guidelines in our [Contributing](https://github.com/Linuxbrew/linuxbrew/blob/master/.github/CONTRIBUTING.md) document?
[x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Linuxbrew/linuxbrew/pulls) for the same update/change?
[x] Have you included a log of the failed build without your patch using `brew gist-logs <formula>`?
[ ] Does your submission pass
[x] Have you built your formula locally prior to submission with `brew install <formula>`?

This file does not pass a strict audit, because it lacks a test, but IMO that should be fixed upstream.

Message:
> compilation terminated.
Makefile:42: recipe for target 'main.o' failed
make: *** [main.o] Error 1
make: *** Waiting for unfinished jobs....
Makefile:42: recipe for target 'state.o' failed
make: *** [state.o] Error 1
Makefile:42: recipe for target 'config.o' failed
make: *** [config.o] Error 1
In file included from helpers.c:15:0:
main.h:80:28: fatal error: machine/endian.h: No such file or directory
 #include <machine/endian.h>
                            ^
compilation terminated.
Makefile:42: recipe for target 'helpers.o' failed
make: *** [helpers.o] Error 1

### DIFF
--- a/Library/Formula/foremost.rb
+++ b/Library/Formula/foremost.rb
@@ -11,7 +11,12 @@ class Foremost < Formula
       s.change_make_var! "RAW_FLAGS", ENV.cflags
     end
 
-    system "make", "mac"
+    if OS.mac?
+      system "make", "mac"
+    else
+      inreplace "config.c", "/usr/local/etc/", "#{etc}/"
+      system "make"
+    end
 
     bin.install "foremost"
     man8.install "foremost.8.gz"


Library/Formula/foremost.rb
```diff
@@ -11,7 +11,12 @@ def install
       s.change_make_var! "RAW_FLAGS", ENV.cflags
     end
 
-    system "make", "mac"
+    if OS.mac?
+      system "make", "mac"
+    else
+      inreplace "config.c", "/usr/local/etc/", "#{etc}/"
```
rwhogg: This line isn't strictly necessary, but it makes the formula respect Linuxbrew conventions better. (And it can't be changed by altering a Makefile variable: it is hard-coded into the source).

Without it, on startup the command tries to look for the default config file in `/usr/local` (which by sheer coincidence is where upstream Homebrew would have put it).

Technically speaking, it *still* throws a warning message on startup because of the rename to `foremost.conf.default`, but given that the formula already did that I assume that was by design.

sjackman:
Thanks once again for your contribution, Bob!

PR #918, opened 2016-03-12
Fix for ARM Linuxbrew for Raspberry Pi

AlexisMeza:
Resolved issues on Raspberry Pi with glibc

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -194,7 +194,9 @@ module Homebrew
     ld_so = HOMEBREW_PREFIX/"lib/ld.so"
     return if ld_so.readable?
     glibc = Formula["glibc"]
-    sys_interpreter = File.exist?("/lib64/ld-linux-x86-64.so.2") ? "/lib64/ld-linux-x86-64.so.2" : "/lib/ld-linux.so.3"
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/install.rb
```diff
@@ -194,7 +194,8 @@ def check_ld_so_symlink
     ld_so = HOMEBREW_PREFIX/"lib/ld.so"
     return if ld_so.readable?
     glibc = Formula["glibc"]
-    interpreter = glibc && glibc.installed? ? glibc.lib/"ld-linux-x86-64.so.2" : "/lib64/ld-linux-x86-64.so.2"
+    sys_interpreter = File.exist?("/lib64/ld-linux-x86-64.so.2") ? "/lib64/ld-linux-x86-64.so.2" : "/lib/ld-linux.so.3"
```
sjackman: Can you change this to: (untested)
```ruby
sys_interpreter = ["/lib64/ld-linux-x86-64.so.2", "/lib/ld-linux.so.3", "/lib/ld-linux.so.2"].find do |s|
        Pathname.new(s).executable?
end
```

sjackman:
Thanks for the patch, Alexis. It's looking good.

AlexisMeza:
I tested your change and it worked.

sjackman:
Merged. Thanks, Alexis!

Issue #899, opened 2016-03-08
libsoup fails to install when XDG_DATA_DIRS is not set to $(brew --prefix)/share


DoomHammer:
Tried upgrading `libsoup` and got greeted with

```
$HOME/.linuxbrew/Cellar/vala/0.30.1/bin/vapigen-0.30 --metadatadir=../libsoup --pkg gio-2.0 --library=libsoup-2.4 Soup-2.4.gir Soup-2.4.metadata Soup-2.4-custom.vala
error: Package `Gio-2.0' not found in specified Vala API directories or GObject-Introspection GIR directories
Generation failed: 1 error(s), 0 warning(s)
make[2]: *** [libsoup-2.4.vapi] Error 1
```

After searching a bit I found out that `vapigen` makes use of `$XDG_DATA_DIRS` but in a curious way.
- If the variable is unset it succeeds.
- If the variable is set and contains `$(brew --prefix)/share` it also succeeds.
- But if the variable is set and does not contain `$(brew --prefix)/share` then it fails as seen above.

Now, in my case `$XDG_DATA_DIRS` is set by my window manager and contains only system paths. I have modified my shell's rc file to prepend `$(brew --prefix)/share` to this variable but this is obviously a workaround.

Question is: should `brew` modify this variable during installation (what about runtime then?) or should user be responsible for setting it correctly in one's own `.rc` (which should then be documented and `brew doctor` should whine about incorrect setting)?

More info on `$XDG_*` can be found in [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

sjackman:
I suppose it depends on how pervasive this problem is. If it affects only this one formula, then it could be hacked around within this one formula, but I'm going to guess that it affects many formula. In which case, the user should probably modify their environment globally. Adding a `brew doctor` check makes sense, I think. If `XDG_DATA_DIRS` is not set, then everything is fine. If it is set, check that it contains `$(brew --prefix)/share`.

DoomHammer:
35 formulae depend on Vala so could be potentially problematic. I see two of them having special treatment, though, `libgweather` and `libsoup`:

```
# ensures that the vala files remain within the keg
inreplace "libsoup/Makefile.in", "VAPIDIR = @VAPIDIR@", "VAPIDIR = @datadir@/vala/vapi"
```

DoomHammer:
[...]

PR #919, opened 2016-03-12
Add $XDG_DATA_DIRS checks.

DoomHammer:
As discussed in #899

### DIFF
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -755,6 +755,25 @@ module Homebrew
         end
       end
 
+
+      def check_xdg_data_dirs
+        share = "#{HOMEBREW_PREFIX}/share"
+        homebrew_in_xdg_data_dirs = true
+        if ENV.key?("XDG_DATA_DIRS")
+          xdg_data_dirs = ENV["XDG_DATA_DIRS"].split(File::PATH_SEPARATOR)
+          xdg_data_dirs.length > 0 && homebrew_in_xdg_data_dirs = xdg_data_dirs.include?(share)
+        end
+        unless homebrew_in_xdg_data_dirs
+          <<-EOS.undent
+          Homebrew's share was not found in your XDG_DATA_DIRS but you have
+          this variable set to include other locations.
+          Some programs like `vapigen` may not work correctly.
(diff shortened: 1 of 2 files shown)


sjackman:
Thanks, Piotr. Does this issue only affect `gir`? I would have thought that it would affect any FreeDesktop application that expected to find its data in `XDG_DATA_DIRS`.

DoomHammer:
Problem is I don't know how many of them actually make use of this variable. Based on that I assumed this is the only one until proven otherwise.

sjackman:
I'd suggest casting the net a little wider. If the environment variable `XDG_DATA_DIRS` is declared, it should include `$HOMEBREW_PREFIX/share`. If it's not declared, everything is fine.

DoomHammer:
Fine for me, I will update the PR.

Library/Homebrew/diagnostic.rb
```diff
@@ -755,6 +755,25 @@ def check_user_path_3
...
+        end
+        unless homebrew_in_xdg_data_dirs
+          <<-EOS.undent
+          Homebrew's share was not found in your XDG_DATA_DIRS but you have
+          this variable set to include other locations.
+          Some programs like `vapigen` may not work correctly.
+          Consider setting the XDG_DATA_DIRS for example like so
+              echo 'export XDG_DATA_DIRS="#{share}:$XDG_DATA_DIRS"' >> #{shell_profile}
+        EOS
+        end
```
sjackman: Just for style, try (untested)
```ruby
if ENV.key?("XDG_DATA_DIRS") &&
    !ENV["XDG_DATA_DIRS"].split(File::PATH_SEPARATOR).include?(HOMEBREW_PREFIX/"share")
  "..."
end
```

DoomHammer: Not so much ;)

Tried to do it in a more compact way, but I think logic won and I lost...

sjackman:
`brew tests` failed: https://travis-ci.org/Linuxbrew/linuxbrew/builds/120258291#L255
I'm getting ready to look at the grand split into `brew` and `homebrew-core`. After fixing the code style and the test, are you ready to merge this PR?

DoomHammer:
Superseded by https://github.com/Linuxbrew/brew/pull/93

PR #924, opened 2016-03-13
mkdocs: Fix for linuxbrew (add resource for setuptools)

hartzell:
The python in sjackman's docker container does not have
setuptools installed, so we'll be safe and pull in our own.

The homebrew folks told me that there is no harm in doing this
(vis a vis six...).

### DIFF
--- a/Library/Formula/mkdocs.rb
+++ b/Library/Formula/mkdocs.rb
@@ -13,6 +13,11 @@ class Mkdocs < Formula
 
   depends_on :python if MacOS.version <= :snow_leopard
 
+  resource "setuptools" do
+    url "https://pypi.python.org/packages/source/s/setuptools/setuptools-20.2.2.tar.gz"
+    sha256 "24fcfc15364a9fe09a220f37d2dcedc849795e3de3e4b393ee988e66a9cbd85a"
+  end
+
   resource "tornado" do
     url "https://pypi.python.org/packages/source/t/tornado/tornado-4.3.tar.gz"
     sha256 "c9c2d32593d16eedf2cec1b6a41893626a2649b40b21ca9c4cac4243bde2efbf"


hartzell:
Looks like the travis test failed trying to build a bottle, for lack of homebrew-dups not having been tapped.

What's my next move?

sjackman:
The `homebrew/dupes` issue seems to be because the dependency is on `depends_on :python` rather than `depends_on "python"`, that is, it's a `Requirement` rather than a `Dependency`. I would call that a bug in how `Requirement` are resolved.

sjackman:
The work around is to `apt-get install python-setuptools` or `brew install python`.

sjackman:
This issue #931 has a similar problem.

sjackman:
I'd prefer to fix this issue by adding a `brew doctor` check that checks that if Python is installed, then `setuptools` is also installed. Either that or change the `python` requirement so that if `setuptools` is not installed, it reports that `python` is not installed.

DoomHammer:
:+1: for the Doctor. :hospital:

sjackman:
@DoomHammer Would you like to stab at implementing the Doctor check? It should be pretty straight forward. See https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/test/test_language_module_requirement.rb#L12
```ruby
LanguageModuleRequirement.new("setuptools").satisfied?
```

hartzell:
@DoomHammer -- please take a stab at this if you have the opportunity.  I still have it on my plate but haven't had the time to act on it.

DoomHammer:
Well, :watch: is rarely on our side. Once I had it I will give this a shot.

sjackman:
@DoomHammer Could you brew doctor check for Python setuptools also check that the Python headers are installed?

DoomHammer:
Got it on my list! There were many times when I was bitten by lack of `python-dev` ;)

sjackman:
The brew doctor output should suggest something like
```
To resolve this issue, either `brew install python` (recommended)
or `sudo apt-get install python-dev python-setuptools` on Debian and Ubuntu
or `sudo yum install STUFF GOES HERE` on Fedora and CentOS
```

sjackman:
Superseded by #973. Thanks for this contribution, George, all the same.

PR #925, opened 2016-03-13
packer: Fix for linuxbrew (set XC_OS from OS::NAME)

hartzell:
This sets XC_OS from OS::NAME instead of hardcoding "darwin".  This makes things work.

This only works if I've installed the python-dev package (via apt-get on an ubuntu 14.04 Digital Ocean droplet).  Without that there's a bit of a rabbit hole that I never figured out as packer needs mercurial which needs python's header files.

### DIFF
--- a/Library/Formula/packer.rb
+++ b/Library/Formula/packer.rb
@@ -354,7 +354,7 @@ class Packer < Formula
   end
 
   def install
-    ENV["XC_OS"] = "darwin"
+    ENV["XC_OS"] = OS::NAME
     ENV["XC_ARCH"] = MacOS.prefer_64_bit? ? "amd64" : "386"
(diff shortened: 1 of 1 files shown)


rwhogg:
IIRC the preferred format for commit messages for these sorts of porting fixes is "<formula name>: Fix for Linuxbrew".

@sjackman, would you accept a PR to say that in CONTRIBUTING.md?

sjackman:
> @sjackman, would you accept a PR to say that in CONTRIBUTING.md?

Yes, absolutely. Thanks, Bob. The preferred short subject line is as above. The rest of the message can be whatever the user wants. Please include a link to http://chris.beams.io/posts/git-commit/

sjackman:
Merged. Thanks for your contribution, George!

Issue #953, opened 2016-03-20
glibc: depends on gawk


DoomHammer:
Building from source because of #952 I got:

```
==> ../configure --disable-debug --disable-dependency-tracking --disable-silent-rules --prefix=/home/linuxbrew/.linuxbrew/Cellar/glibc/2.19 --enable-obsolete-rpc --without-selinu
Last 15 lines from /home/linuxbrew/.cache/Homebrew/Logs/glibc/01.configure:
checking for gnumake... no
checking for gmake... no
checking for make... make
checking version of make... 3.81, ok
checking for gnumsgfmt... no
checking for gmsgfmt... no
checking for msgfmt... no
checking for makeinfo... no
checking for sed... sed
checking version of sed... 4.2.2, ok
checking for gawk... no
checking for nm... nm
configure: error: 
*** These critical programs are missing or too old: gawk
*** Check the INSTALL file for required versions.
```

Doing further tests whether simply adding a dependency will solve the issue.

Env: Docker `linuxbrew/linuxbrew` `7c827ee662db`, Homebrew `f0e235a`

rwhogg:
At https://sourceware.org/glibc/wiki/FAQ#Compiling_glibc, there's a list of necessary build dependencies, including `gawk` and `perl`.

Thanks for the bug report @DoomHammer!

DoomHammer:
Fascinating thing is that installing deps one by one seems to work. Only trying to `brew` something that installs a dep first fails.

Separate issue: `glibc` won't build with `gcc-4.3`. Seems the :chicken: is eating its :egg: or something alike... Catch 22?

sjackman:
Yes, even CentOS 5 has `gcc 4.4` available as an optional package, which is able to build `glibc`. I'm not sure how to install `glibc` with a system that has `gcc 4.3`. Install a deb for a newer release of GCC if you can. Perhaps a backport?
It might be possible to do something like
```
brew install gcc --without-glibc
brew install glibc
brew install gcc --with-glibc
```
but it would be tricky.

sjackman:
Currently the `README.md` says that the minimum version off GCC is `4.2`. I'm going to increase that to `4.4`.

DoomHammer:
Pieces don't fit. `gcc-4.4` in Debian pulls in `glibc` 2.11 so back to the drawing board.

sjackman:
[...]

PR #955, opened 2016-03-20
glibc: depends on gawk

DoomHammer:
Details: #953

### DIFF
--- a/Library/Formula/glibc.rb
+++ b/Library/Formula/glibc.rb
@@ -16,6 +16,8 @@ class Glibc < Formula
   # binutils 2.20 or later is required
   depends_on "binutils" => [:build, :recommended]
 
+  depends_on "gawk" => :build
+
   # Linux kernel headers 2.6.19 or later are required
   depends_on "linux-headers" => [:build, :recommended]
 


sjackman:
`xz` was installed by bottle, but that bottle is not working on the Travis CI system.
https://travis-ci.org/Linuxbrew/linuxbrew/builds/117300152#L655-L658
```
==> Downloading http://ftpmirror.gnu.org/gawk/gawk-4.1.3.tar.xz
Already downloaded: /home/travis/.cache/Homebrew/gawk-4.1.3.tar.xz
==> Verifying gawk-4.1.3.tar.xz checksum
Error: Empty archive
```

sjackman:
Travis CI is running Ubuntu 12 Trusty.
It compiled successfully on LinuxbrewTestBot running Ubuntu 14 Precise.
See https://github.com/LinuxbrewTestBot/linuxbrew/tree/pr-DoomHammer-add-glibc-deps

sjackman:
See #968 xz: Error: Empty archive

DoomHammer:
Yet another reason to give #601 a try? ;)

sjackman:
I fixed #968 xz: Error: Empty archive

sjackman:
I would like any deps of the toolchain (i.e. `glibc` and `gcc`) to be bottled, to make installing the toolchain quick and painless. `gawk` doesn't say it's `cellar :any`, but I suspect it would work just fine.

Anything bottles installed before `glibc` (that is any deps of `glibc`, like `zlib` and `binutils`) have to be built on a "holy build box" with glibc 2.5, the oldest glibc supported to bootstrap Linuxbrew. The downside to that solution is that these "compatible" bottles have to be built by hand. So I want to avoid adding hard deps to `glibc`, and most systems do provide `gawk`.

I suggest adding a `Requirement` rather than a dependency that first checks whether the system provides `awk`, and if not builds `gawk` from source. `gawk` takes one minute to build, so perhaps I'm overthinking this, but it's a simple fix.

```ruby
class AwkRequirement < Requirement
  fatal true
  default_formula "gawk"
  satisfy { which "awk" }
end
```

DoomHammer:
Understood, sir!

BTW: why build by hand on a holy machine? Can't we make a holy Docker container just for that?

sjackman:
For sure! Any distribution that uses `glibc` 2.5 will do just fine. CentOS 5 for example, though I'd prefer a Debian or Ubuntu release that also used `glibc` 2.5.

DoomHammer:
BTW: where does `linuxbrew/linuxbrew` come from as it is not the `Dockerfile` from current repo for sure?

DoomHammer:
OK, nevermind that. I pushed `doomhammer/linuxbrew-lenny` to Docker cloud but wouldn't call it working.

When I issued `brew install gcc --with-glibc --verbose` it got stuck:

```
==> Installing dependencies for gcc: patchelf, zlib, binutils, linux-headers, glibc, xz, gmp, gpatch, mpfr, libmpc, isl
==> Installing gcc dependency: patchelf
==> Downloading https://linuxbrew.bintray.com/bottles/patchelf-0.9.x86_64_linux.bottle.tar.gz
/usr/bin/curl -fLA Homebrew 0.9.5 (Ruby 2.1.5-273; x86_64-linux) https://linuxbrew.bintray.com/bottles/patchelf-0.9.x86_64_linux.bottle.tar.gz -C 0 -o /home/linuxbrew/.cache/Home
brew/patchelf-0.9.x86_64_linux.bottle.tar.gz.incomplete
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  285k  100  285k    0     0   129k      0  0:00:02  0:00:02 --:--:--  140k
==> Verifying patchelf-0.9.x86_64_linux.bottle.tar.gz checksum
==> Pouring patchelf-0.9.x86_64_linux.bottle.tar.gz
tar xvf /home/linuxbrew/.cache/Homebrew/patchelf-0.9.x86_64_linux.bottle.tar.gz
tar: Failed to set default locale
x patchelf/0.9/
x patchelf/0.9/bin/
x patchelf/0.9/bin/patchelf
x patchelf/0.9/share/
x patchelf/0.9/share/doc/
x patchelf/0.9/share/doc/patchelf/
x patchelf/0.9/share/doc/patchelf/README
x patchelf/0.9/share/man/
x patchelf/0.9/share/man/man1/
x patchelf/0.9/share/man/man1/patchelf.1
x patchelf/0.9/COPYING
x patchelf/0.9/INSTALL_RECEIPT.json
x patchelf/0.9/README
```

Funny thing, running `brew install patchelf` seemed to work, but then:

```
==> Installing dependencies for gcc: zlib, binutils, linux-headers, glibc, xz, gmp, gpatch, mpfr, libmpc, isl
==> Installing gcc dependency: zlib
==> Downloading https://linuxbrew.bintray.com/bottles/zlib-1.2.8.x86_64_linux.bottle.1.tar.gz
Already downloaded: /home/linuxbrew/.cache/Homebrew/zlib-1.2.8.x86_64_linux.bottle.1.tar.gz
==> Verifying zlib-1.2.8.x86_64_linux.bottle.1.tar.gz checksum
==> Pouring zlib-1.2.8.x86_64_linux.bottle.1.tar.gz
tar xvf /home/linuxbrew/.cache/Homebrew/zlib-1.2.8.x86_64_linux.bottle.1.tar.gz
tar: Failed to set default locale
x zlib/1.2.8/
x zlib/1.2.8/ChangeLog
x zlib/1.2.8/lib/
x zlib/1.2.8/lib/libz.so.1
x zlib/1.2.8/lib/libz.so.1.2.8
x zlib/1.2.8/lib/pkgconfig/
x zlib/1.2.8/lib/pkgconfig/zlib.pc
x zlib/1.2.8/lib/libz.a
x zlib/1.2.8/lib/libz.so
x zlib/1.2.8/README
x zlib/1.2.8/INSTALL_RECEIPT.json
x zlib/1.2.8/include/
x zlib/1.2.8/include/zlib.h
x zlib/1.2.8/include/zconf.h
x zlib/1.2.8/share/
x zlib/1.2.8/share/man/
x zlib/1.2.8/share/man/man3/
x zlib/1.2.8/share/man/man3/zlib.3
/home/linuxbrew/.linuxbrew/Cellar/patchelf/0.9/bin/patchelf --set-rpath /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/Cellar/zlib/1.2.8/lib/libz.so.1.2.8
```

Could you point me @sjackman where to go from here as I am still not very familiar with Homebrew internals?

Thanks!

sjackman:
`tar: Failed to set default locale`
See https://github.com/Linuxbrew/linuxbrew/blob/master/Dockerfile#L9

sjackman:
Is it just hanging after extracting the tarball? I've seen that behaviour too on Ubuntu 12 precise, and haven't figured it out.

sjackman:
For my own reference:
History of Debian releases: https://wiki.debian.org/DebianReleases
History of glibc releases: https://sourceware.org/glibc/wiki/Glibc%20Timeline

DoomHammer:
`localedef` on lenny have no notion of `-i` switch. That is why I omitted
this.

As for hanging: this is what it looks like. May try a bit of `strace`ing
but other than that, no idea.

sjackman:
Lenny probably didn't support UTF-8 either.

I noticed that Ruby was still using CPU, so I think it was stuck in some kind of loop. Possibly (long shot here) due to a missing program.

DoomHammer:
I'd also bet for a missing program. Maybe `command-not-found`'s approach would help here?

DoomHammer:
Funny thing. Should `cc1` be running during pouring bottle? I thought this process was more like: "unpack your things and move them to correct places"...

sjackman:
No, `cc1` should not be running when pouring a bottle. `pstree` may be useful to see who's running it.

DoomHammer:
While running `brew install binutils --verbose`:

```
$ pstree
bash-+-pstree
     `-ruby-+-ruby-+-ruby-+-make---sh---make---sh---sh---make---make---sh---make---8*[gcc-4.3-+-as]
            |      |      |                                                                   `-cc1]
            |      |      `-{ruby}
            |      `-{ruby}
            `-{ruby}
```

The output on the other hand was:

```
[...]
/home/linuxbrew/.linuxbrew/Cellar/patchelf/0.9/bin/patchelf --set-rpath /home/linuxbrew/.linuxbrew/lib --set-interpreter /home/linuxbrew/.linuxbrew/lib/ld.so /home/linuxbrew/.linuxbrew/Cellar/binutils/2.25.1/x86
_64-unknown-linux-gnu/bin/objdump 
/home/linuxbrew/.linuxbrew/Cellar/patchelf/0.9/bin/patchelf --set-rpath /home/linuxbrew/.linuxbrew/lib --set-interpreter /home/linuxbrew/.linuxbrew/lib/ld.so /home/linuxbrew/.linuxbrew/Cellar/binutils/2.25.1/x86
_64-unknown-linux-gnu/bin/ranlib 
/home/linuxbrew/.linuxbrew/Cellar/patchelf/0.9/bin/patchelf --set-rpath /home/linuxbrew/.linuxbrew/lib --set-interpreter /home/linuxbrew/.linuxbrew/lib/ld.so /home/linuxbrew/.linuxbrew/Cellar/binutils/2.25.1/x86
_64-unknown-linux-gnu/bin/strip
```

DoomHammer:
Come to think of it, maybe 4.3 was deprecated based on fact it has some weird loop issue or some such?

Still: what is that guy even doing when I am pouring from a bottle?

sjackman:
It's clearly compiling something. It looks as though the output is not updated, but brew has moved on to compiling a formula. Does `ps aux |egrep 'make|sh'` give some indication of what's being compiled? You can also poke around in `/proc` to figure out what these processes are up to.

DoomHammer:
What is being compiled is usually what is being installed. I am more concerned whether it is done after pouring. Is there some logic in place to build from source after pouring is done (eg. if poured bottled is incompatible with the system or something like that)?

DoomHammer:
Another interesting question: what could be blocking output? I clearly see in `ps aux` that installation of `binutils` is taking place whereas console seems stuck on:

```
brew install binutils
==> Installing dependencies for binutils: patchelf, zlib
==> Installing binutils dependency: patchelf
==> Downloading https://linuxbrew.bintray.com/bottles/patchelf-0.9.x86_64_linux.bottle.tar.gz
######################################################################## 100.0%
==> Pouring patchelf-0.9.x86_64_linux.bottle.tar.gz
```

DoomHammer:
In the meanwhile `zlib` got silently installed as well.

sjackman:
It looks like the console has hang up. Perhaps its file descriptor was inadvertently closed.

DoomHammer:
Tried my luck with Centos and things were mostly smooth, until `gawk` couldn't link. Maybe this is a dead end, in fact?

As a workaround one might create a fake keg with `curl -Lq s.minos.io/s > /home/linuxbrew/static-get &&         sh /home/linuxbrew/static-get -x gawk`...

sjackman:
That's the first that I've heard of `minos`. It's not a bad workaround, but I'd prefer to install `gawk` using Linuxbrew.

DoomHammer:
I would too, but no luck on ancient distro ;)

sjackman:
Where are you stuck? With the broken console? I haven't figured that one out yet myself.
I regularly use Linuxbrew on CentOS 5.11 with glibc 2.5.

DoomHammer:
Console problems were on Debian Lenny. `gawk` on CentOS couldn't build due to some reloc problem, which I suspect means linker is too old.

sjackman:
Can you post the error message?

DoomHammer:
My bad, linking error was with something different altogether. `gawk` was pretty simple:

```
==> Installing dependencies for gawk: xz
==> Installing gawk dependency: xz
==> Downloading https://linuxbrew.bintray.com/bottles/xz-5.2.2.x86_64_linux.bottle.tar.gz
######################################################################## 100.0%
==> Pouring xz-5.2.2.x86_64_linux.bottle.tar.gz
/home/linuxbrew/.linuxbrew/Cellar/xz/5.2.2: 65 files, 1.6M
==> Installing gawk
==> Downloading http://ftpmirror.gnu.org/gawk/gawk-4.1.3.tar.xz
==> Downloading from http://ftp.task.gda.pl/pub/gnu/gawk/gawk-4.1.3.tar.xz
######################################################################## 100.0%
Error: Empty archive
```

This is both well-known and a circular dependency (I cannot install `glibc` if we want `gawk` linked against the old `glibc`). Workaround might perhaps be to install non-brewed `xz`?

sjackman:
Ah, `xz` is linked against new `glibc`. As a temporary workaround, use the `gz` `http://ftpmirror.gnu.org/gawk/gawk-4.1.3.tar.gz` rather than `xz`.

sjackman:
As discussed, this fix requires an `:awk` requirement, so as to avoid adding a build dependency to `binutils`. Closing this PR for now, before the grand split of `brew` and `homebrew-core`.

Issue #946, opened 2016-03-17
pam: Create formula for Linuxbrew


sjackman:
Homepage: http://www.linux-pam.org/

sjackman:
rstudio-server: Depends on pam #819

rwhogg:
I'm happy to take this one on, but it's not all that high priority for me right now. I'll increase priority on it if you want to use R Studio, but since the only other formula I'm aware of that uses it is CVS (not exactly the most important formula in the world these days) I think dealing with X11 is more important right now.

sjackman:
Agreed. I use RStudio on Mac. On Linux I use only R command line.

rwhogg:
This issue was moved to Linuxbrew/homebrew-core#131

PR #969, opened 2016-03-22
cvs: Fix for Linuxbrew

rwhogg:
There were two problems with `cvs`:

1\. It depends on `vim`, which isn't necessarily installed by default on Linux (Slackware uses `elvis` instead, but I noticed because I actually manually removed it from my Ubuntu system!)

```
checking for vim... no
configure: error:
   Failed to find a text file editor.  CVS cannot be compiled
   without a default log message editor.  Searched for
   vim'.  Try configure --with-editor'.
```

2\. There was a patch which added an unsupported `gcc` flag.

```
Makefile:572: recipe for target 'admin.o' failed
make[2]: *** [admin.o] Error 1
gcc-5: error: unrecognized command line option '-mdynamic-no-pic'
gcc-5: error: unrecognized command line option '-mdynamic-no-pic'
Makefile:572: recipe for target 'annotate.o' failed
```

### DIFF
--- a/Library/Formula/cvs.rb
+++ b/Library/Formula/cvs.rb
@@ -3,6 +3,12 @@
 # MacPorts: https://trac.macports.org/browser/trunk/dports/devel/cvs/Portfile
 # Creating a useful testcase: http://mrsrl.stanford.edu/~brian/cvstutorial/
 
+class VimRequirement < Requirement
+  fatal true
+  default_formula "vim"
+  satisfy { which "vim" }
+end
+
 class Cvs < Formula
   desc "Version control system"
   homepage "http://cvs.nongnu.org/"
@@ -20,21 +26,23 @@ class Cvs < Formula
 
   keg_only :provided_until_xcode5
 
+  depends_on VimRequirement
+
   patch :p0 do
     url "https://opensource.apple.com/tarballs/cvs/cvs-45.tar.gz"
     sha256 "4d200dcf0c9d5044d85d850948c88a07de83aeded5e14fa1df332737d72dc9ce"
-    apply "patches/PR5178707.diff",
+    apply *["patches/PR5178707.diff",
           "patches/ea.diff",
           "patches/endian.diff",
           "patches/fixtest-client-20.diff",
           "patches/fixtest-recase.diff",
           "patches/i18n.diff",
           "patches/initgroups.diff",
-          "patches/nopic.diff",
+          ("patches/nopic.diff" if OS.mac?),
           "patches/remove-libcrypto.diff",
           "patches/remove-info.diff",
           "patches/tag.diff",
-          "patches/zlib.diff"
+          "patches/zlib.diff"].compact
   end
 
   def install


sjackman:
> It depends on `vim`

That is shocking. Oi. I'd rather not make this a hard dependency, since most systems provided `vi`. Can you make this dep a `Requirement` instead? See #953 

```
class ViRequirement < Requirement
  fatal true
  default_formula "vim"
  satisfy { which "vi" }
end
```

sjackman:
Try to avoid changing the formatting where possible. Large changes like this tend to create merge conflicts when I do the weekly merge of Homebrew into Linuxbrew. A one-line change is always preferred where possible. Try something like
```ruby
    apply "patches/PR5178707.diff",
          "patches/ea.diff",
          "patches/endian.diff",
          "patches/fixtest-client-20.diff",
          "patches/fixtest-recase.diff",
          "patches/i18n.diff",
          "patches/initgroups.diff",
          ("patches/nopic.diff" if OS.mac?),
          "patches/remove-libcrypto.diff",
          "patches/remove-info.diff",
          "patches/tag.diff",
          "patches/zlib.diff"
```
If `apply` doesn't like the `nil` value, it could be removed with `compact`, but I'd suggest patching `apply` instead to ignore `nil` values.

rwhogg:
> If apply doesn't like the nil value, it could be removed with compact, but I'd suggest patching apply instead to ignore nil values.

IMO, patching `apply` is something that should probably happen upstream (unless we do it for just the `cvs` formula, but that sounds heavy-handed to me). I think the path of least resistance in this case is probably `compact`.

Nevertheless, I'll try to avoid invasive changes to the formatting in the future. Thanks for the help while I get used to this!

sjackman:
Let's use `compact` here then. If a similar case comes up again in a different formula, I'd be inclined to patch `apply`. It would be nice for it to be merged upstream, but they'd be unlikely to accept it, because the issue of `nil` arguments being passed to `apply` doesn't affect Mac.

rwhogg:
There was a similar case before - for sip.

(No pr because im on my Android phone, sorry)

On March 23, 2016 10:15:04 AM PDT, Shaun Jackman <notifications@github.com> wrote:
>Let's use `compact` here then. If a similar case comes up again in a
>different formula, I'd be inclined to patch `apply`. It would be nice
>for it to be merged upstream, but they'd be unlikely to accept it,
>because the issue of `nil` arguments being passed to `apply` doesn't
>affect Mac.
>
>---
>You are receiving this because you were assigned.
>https://github.com/Linuxbrew/linuxbrew/pull/969#issuecomment-200447637

-- 

sjackman:
https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Formula/sip.rb#L35
```
      system python, *["configure.py",
                     ("--deployment-target=#{MacOS.version}" if OS.mac?),
                     "--destdir=#{lib}/python#{version}/site-packages",
                     "--bindir=#{bin}",
                     "--incdir=#{include}",
                     "--sipdir=#{HOMEBREW_PREFIX}/share/sip"].compact
```
Similar yes. I'm not suggesting that we monkey patch `system` though. Now that I say it though… it would actually get rid of a lot of these somewhat ugly calls to `compact`. Seems a bit drastic though.
Passing `nil` values to `system configure` is fine, the `nil` values get converted to empty strings `""` and `configure` just ignores the empty strings. Passing empty strings to `make` though makes it unhappy.

sjackman:
LinuxbrewTestBot:
```
checking security/pam_appl.h usability... no
checking security/pam_appl.h presence... no
checking for security/pam_appl.h... no
checking pam/pam_appl.h usability... no
checking pam/pam_appl.h presence... no
checking for pam/pam_appl.h... no
configure: error: Could not find PAM headers
```
How do we install PAM on Linuxbrew?

sjackman:
`"--enable-pam",`
or should we disable PAM for Linuxbrew?

rwhogg:
I think issue #946 was to add a `pam` formula. That might do it.

rwhogg:
Alternatively, we could disable pam as you suggest.

sjackman:
I'm fine with either solution (if you're volunteering for #946). =)

sjackman:
I'm happy to merge this PR as is. We can tackle the PAM dependency another day.

rwhogg:
Cool. I'll merge it later tonight.

PR #970, opened 2016-03-22
ghc: Depends on m4 for Linuxbrew

sjackman:
Fix error:
`checking for suitable m4... configure: error: No usable m4 in $PATH`

### DIFF
--- a/Library/Formula/ghc.rb
+++ b/Library/Formula/ghc.rb
@@ -14,7 +14,8 @@ class Ghc < Formula
   deprecated_option "tests" => "with-test"
   deprecated_option "with-tests" => "with-test"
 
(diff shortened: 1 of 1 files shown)


sjackman:
```
configure: WARNING: unrecognized options: --with-ld
checking for path to top of build tree... /tmp/ghc--binary20160322-22146-nqdho3/ghc-7.6.3
ghc/stage2/build/tmp/ghc-stage2: error while loading shared libraries: libncursesw.so.5: cannot open shared object file: No such file or directory
Can't work out build platform
```
https://travis-ci.org/Linuxbrew/linuxbrew/builds/117644755#L1177-L1180

PR #971, opened 2016-03-22
attr: install headers

DoomHammer:
Other formulae that depend on attr cannot be properly compiled if headers are not installed.


sjackman:
Thanks, Piotr. I'll merge this.

Could you submit a PR to fix this issue from Docker Cloud as well, please?
```
checking for msgfmt... no
FATAL ERROR: msgfmt does not seem to be installed.
attr cannot be built without a working gettext installation.
```

DoomHammer:
This happens when installing `attr`?

sjackman:
Yes.

Issue #632, opened 2015-11-06
gtk+ failed to build on linuxbrew standalone Ubuntu 14.04


arunnair411:
As part of getting opencv3 up and running, I was trying to install gtk+ through linuxbrew. It includes cairo as a dependency, which it installs displaying

```
==> ./configure --prefix=/home/arun/.linuxbrew/Cellar/cairo/1.14.2_1 --enable-gobject=yes --enable-svg=yes --enable-tee=yes --enable-xcb=no --enable-xlib=no --enable-xlib-xrender=no
```

Looking at cairo.rb [at](https://github.com/Homebrew/linuxbrew/blob/master/Library/Formula/cairo.rb), I see that enable-xlib is set to no as the x11 dependency is not satisfied. However, later in the installation process, 

```
configure: error: Package requirements (cairo-xlib >= 1.6) were not met:

No package 'cairo-xlib' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables CAIRO_BACKEND_CFLAGS
and CAIRO_BACKEND_LIBS to avoid the need to call
```

which seems to imply that enable-xlib should be true. How do I fix this? Here is the [gist log](https://gist.github.com/anonymous/57ef49a8346d46408fec)

arunnair411:
Any thoughts as to what I can do to get it working?

ahundt:
I don't think the linuxbrew x11 is what should be used here either, perhaps it should be the one provided by the system, even in the standalone case? Can x11 even be compiled and run properly in a truly standalone way? I feel like at some level dependencies must exist to the underlying system.

DoomHammer:
Running `brew install cairo --with-x11` solves the issue. I do have, however `linuxbrew/xorg/xorg` installed locally.

sjackman:
See also #542.
Fixed by #974.

PR #974, opened 2016-03-22
gtk+: depends on cairo --with-x11 for Linuxbrew

DoomHammer:
As reported in #632

### DIFF
--- a/Library/Formula/gtk+.rb
+++ b/Library/Formula/gtk+.rb
@@ -29,6 +29,7 @@ class Gtkx < Formula
   depends_on "gdk-pixbuf"
   depends_on "jasper" => :optional
   depends_on "atk"
+  depends_on "cairo" => "with-x11" unless OS.mac?
   depends_on "pango"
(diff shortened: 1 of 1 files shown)


sjackman:
:+1+

sjackman:
Merged. Thanks, Piotr!

sjackman:
On second thought, this issue should have been resolved differently.
`:x11` should be a `:recommended` dependency of `cairo` for non-Mac systems.
Would you like to take a stab at a patch that does that, then I'll revert this patch.

DoomHammer:
Not sure if I am capable but will give this a try tomorrow (so probably late evening for you, @sjackman ;) ).

sjackman:
See for example https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Formula/htop.rb#L18

DoomHammer:
Something like #986?

sjackman:
Yep, that looks about right to me. Thanks, Piotr.

PR #976, opened 2016-03-23
libpng: Depends on zlib for Linuxbrew

sjackman:
Fix configure: error: zlib not installed

### DIFF
--- a/Library/Formula/libpng.rb
+++ b/Library/Formula/libpng.rb
@@ -24,6 +24,8 @@ class Libpng < Formula
 
   option :universal
 
+  depends_on "zlib" unless OS.mac?
+
   def install
     ENV.universal_binary if build.universal?
     system "./configure", "--disable-dependency-tracking",


sjackman:
```
==> brew test --verbose libpng
Testing libpng
==> /usr/bin/gcc test.c -I/home/travis/build/Linuxbrew/linuxbrew/Cellar/libpng/1.6.21/include -L/home/travis/build/Linuxbrew/linuxbrew/Cellar/libpng/1.6.21/lib -lpng -o test
.
==> ./test
Last 150 lines from /home/travis/.cache/Homebrew/Logs/libpng/02.test:
2016-03-23 00:20:04 +0000
./test
Error: libpng: failed
Failed executing: ./test 
```
https://travis-ci.org/Linuxbrew/linuxbrew/builds/117854580#L1208-L1220

See https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Formula/libpng.rb#L49
```ruby
system ENV.cc, "test.c", "-I#{include}", "-L#{lib}", "-lpng", "-o", "test"
```
`ENV.cc` is not set correctly in the `test` block. It has the value `/usr/bin/gcc`, when it should be `gcc-5`.
`brew test libpng --cc=gcc-5` works.

PR #983, opened 2016-03-24
attr: depends on gettext

DoomHammer:
Fixes #971 Docker Cloud errors.


sjackman:
Merged. Thanks, Piotr!

PR #984, opened 2016-03-25
zsh: Depends on ncurses for Linuxbrew

rwhogg:
`zsh` depends on `ncurses`.

```
configure: error: in `/home/bob/tmp/zsh20160324-12221-1iabe4a/zsh-5.2':
configure: error: "No terminal handling library was found on your system.
This is probably a library called 'curses' or 'ncurses'.  You may
need to install a package called 'curses-devel' or 'ncurses-devel' on your
system."
```

There are unresolved issues building `ncurses` on my machine at the moment; I'm going to leave this PR open until those are resolved.

Similarly, `zsh` doesn't pass a strict audit, but that should be fixed upstream.

### DIFF
--- a/Library/Formula/zsh.rb
+++ b/Library/Formula/zsh.rb
@@ -23,6 +23,7 @@ class Zsh < Formula
   depends_on "gdbm"
   depends_on "pcre"
   depends_on "texinfo" unless OS.mac?
+  depends_on "homebrew/dupes/ncurses" unless OS.mac?
 
   def install
     system "Util/preconfig" if build.head?


sjackman:
Another `zsh` error from LinuxbrewTestBot:
```
** Cannot --enable-pcregrep-libz because zlib.h was not found
```

sjackman:
Ach, that's totally my fault for not giving enough log context. Sorry, Bob.

It was `pcre` that failed to build, not `zlib`. So you can remove the `zlib` dependency from here, and add it to `pcre`.

```
** Cannot --enable-pcregrep-libz because zlib.h was not found
==> Formula
Path: /home/linuxbrew/.linuxbrew/Library/Formula/pcre.rb
```

PR #988, opened 2016-03-26
jsvc: Fix for Linuxbrew

rwhogg:
`jsvc` was using Mac-specific paths and configure flags.

See [https://gist.github.com/rwhogg/544f7f2c0659ad40d5c2#file-01-configure-L27]
> configure: error: /System/Library/Frameworks/JavaVM.framework is not a directory

Strict audit fails because of missing test.

### DIFF
--- a/Library/Formula/jsvc.rb
+++ b/Library/Formula/jsvc.rb
@@ -11,15 +11,17 @@ class Jsvc < Formula
   depends_on :java
 
   def install
-    ENV.append "CFLAGS", "-arch #{MacOS.preferred_arch}"
-    ENV.append "LDFLAGS", "-arch #{MacOS.preferred_arch}"
-    ENV.append "CPPFLAGS", "-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers"
+    if OS.mac?
+      ENV.append "CFLAGS", "-arch #{MacOS.preferred_arch}"
+      ENV.append "LDFLAGS", "-arch #{MacOS.preferred_arch}"
+      ENV.append "CPPFLAGS", "-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers"
+    end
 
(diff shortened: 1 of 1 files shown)


rwhogg:
Travis is failing because of `gcc` (https://travis-ci.org/Linuxbrew/linuxbrew/builds/118636280#L1055): 
> https://travis-ci.org/Linuxbrew/linuxbrew/builds/118636280#L1055

Library/Formula/jsvc.rb
```diff
@@ -11,15 +11,15 @@ class Jsvc < Formula
   depends_on :java
 
   def install
-    ENV.append "CFLAGS", "-arch #{MacOS.preferred_arch}"
-    ENV.append "LDFLAGS", "-arch #{MacOS.preferred_arch}"
-    ENV.append "CPPFLAGS", "-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers"
+    ENV.append "CFLAGS", "-arch #{MacOS.preferred_arch}" if OS.mac?
+    ENV.append "LDFLAGS", "-arch #{MacOS.preferred_arch}" if OS.mac?
+    ENV.append "CPPFLAGS", "-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers" if OS.mac?
```
sjackman: In this I would tend to use an `if` block rather than suffix style `if`.
```ruby
if OS.mac?
    ENV.append "CFLAGS", "-arch #{MacOS.preferred_arch}"
    ENV.append "LDFLAGS", "-arch #{MacOS.preferred_arch}"
    ENV.append "CPPFLAGS", "-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers"
end
```

sjackman: Each line has to be modified anyway, so there's no harm in also changing the formatting.

rwhogg:
Note to self: make a separate PR to bottle this.

PR #991, opened 2016-03-27
node: Build a bottle for Linuxbrew. Do not commit

sjackman:


sjackman:
https://travis-ci.org/Linuxbrew/linuxbrew/builds/118845472#L424-L431
```
==> brew install --only-dependencies --verbose --build-bottle node
Error: No available formula with the name "homebrew/dupes/ncurses" 
Please tap it and then try again: brew tap homebrew/dupes
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
==> FAILED
```
I thought that this bug was fixed. Grumble.

Issue #987, opened 2016-03-25
[WIP] d-bus requires launchd which is unavailable on non-OS X systems


DoomHammer:
```
==> ./configure --prefix=/home/linuxbrew/.linuxbrew/Cellar/d-bus/1.10.6 --localstatedir=/home/linuxbrew/.linuxbrew/var --sysconfdir=/home/linuxbrew/.linuxbrew/etc --disable-xml-docs --disable-doxygen-docs --enab
Last 15 lines from /home/linuxbrew/.cache/Homebrew/Logs/d-bus/01.configure:
configure: WARNING: Sufficiently new AppArmor library not found
checking sys/inotify.h usability... yes
checking sys/inotify.h presence... yes
checking for sys/inotify.h... yes
checking for inotify_init1... yes
checking for Linux epoll(4)... yes
checking sys/event.h usability... no
checking sys/event.h presence... no
checking for sys/event.h... no
checking for kqueue... no
checking launch.h usability... no
checking launch.h presence... no
checking for launch.h... no
checking for launchctl... no
configure: error: launchd support explicitly enabled but not available
```

PR #992, opened 2016-03-28
d-bus: Fix for Linuxbrew

rwhogg:
`d-bus` was using `launchd`, but that's only available on OS X.

Closes Linuxbrew/linuxbrew#987 .

From the issue:
> configure: error: launchd support explicitly enabled but not available

### DIFF
--- a/Library/Formula/d-bus.rb
+++ b/Library/Formula/d-bus.rb
@@ -21,6 +21,8 @@ class DBus < Formula
     sha256 "a8aa6fe3f2d8f873ad3f683013491f5362d551bf5d4c3b469f1efbc5459a20dc"
   end
 
+  depends_on "expat" unless OS.mac?
+
   def install
     # Fix the TMPDIR to one D-Bus doesn't reject due to odd symbols
     ENV["TMPDIR"] = "/tmp"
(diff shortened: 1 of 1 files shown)


DoomHammer:
:+1:

rwhogg:
This formula is missing a dependency on `expat`. I might send a patch for that upstream; we'll see if that's necessary.

DoomHammer:
Isn't `expat` one of those guys that come with OS X?

rwhogg:
@DoomHammer yes it is. (I don't really know what libraries OS X provides, and I was guessing it wasn't provided since `expat` isn't included in homebrew/dupes.). I just confirmed that the missing dependency does not affect OS X.

The dependency is only needed here; I'll add that. No upstream PR necessary.

Library/Formula/d-bus.rb
```diff
@@ -12,6 +12,8 @@ class DBus < Formula
     sha256 "08094b506bf81ce24aa9c38d9414b43726b35d19461440ecfc74d47d02298d50" => :mavericks
   end
 
+  depends_on "expat" unless OS.mac?
```
rwhogg: `expat` isn't included in homebrew/dupes, but it's provided by OS X nonetheless, which is why this formula was missing the dependency.

sjackman:
Some formula that are provided by Mac OS are in the core repository with this tag:
```ruby
keg_only :provided_by_osx, "OS X includes Expat 1.5."
```
https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Formula/expat.rb#L21

PR #993, opened 2016-03-28
Merge 20160328

sjackman:
Merge branch homebrew/master into linuxbrew/master
    
Conflicts:
        Library/Formula/android-ndk.rb
        Library/Formula/app-engine-go-64.rb
        Library/Formula/binutils.rb
        Library/Formula/devil.rb
        Library/Formula/node.rb


sjackman:
Merged.

PR #1013, opened 2016-03-30
Merge 20160330 2

sjackman:


sjackman:
See #1016 (liblacewing) and #1018 (rtags)

Issue #1008, opened 2016-03-30
llvm: Depends on libxml2


dragonxlwang:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

_You can erase any parts of this template not applicable to your Issue._

### Bug reports:

```sh
brew install llvm --with-clang
```
failes with fresh install of brew on Scientific Linux release 6.7 (Carbon)

here's the [log](https://gist.github.com/21431926f23f231c720720be491da9fb)

It reports
```
/home/xwang95/.linuxbrew/bin/ar: ../../libLLVMR600CodeGen.a: No space left on device
make[2]: *** [lib/libLLVMR600CodeGen.a] Error 1
make[2]: *** Deleting file `lib/libLLVMR600CodeGen.a'
make[2]: Leaving directory `/tmp/llvm20160330-27652-1evsf8t'
make[1]: *** [lib/Target/R600/CMakeFiles/LLVMR600CodeGen.dir/all] Error 2
collect2: error: ld returned 1 exit status
```

which I don't quite understand as it appears that on the system I still have plenty of space left (1.7T)

sjackman:
It builds in a temporary directory `/tmp` by default, which may be full, as it's often configured as a small and fast RAM disk. Set the environment variable `HOMEBREW_TEMP`
```sh
mkdir ~/tmp
export HOMEBREW_TEMP=~/tmp
brew install llvm --with-clang
```

dragonxlwang:
I tried the walkaround and it still fails.

This time the error seems related to LD:

```
/home/xwang95/.linuxbrew/bin/ld: cannot find -lxml2
collect2: error: ld returned 1 exit status
make[2]: *** [bin/c-index-test] Error 1
make[2]: Leaving directory `/home/xwang95/tmp/llvm20160330-19540-1rm3q3o'
make[1]: *** [tools/clang/tools/c-index-test/CMakeFiles/c-index-test.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[2]: Leaving directory `/home/xwang95/tmp/llvm20160330-19540-1rm3q3o'
```
And the [gist](https://gist.github.com/dragonxlwang/1b71836ea4f24f9dbd2893e77d108a10)

sjackman:
Try `brew install xml2 llvm --with-clang`

[...]

PR #1014, opened 2016-03-30
llvm: Depends on libxml2 for Linuxbrew

sjackman:
Fix the error `ld: cannot find -lxml2`

Closes Linuxbrew/linuxbrew#1008.


sjackman:
`The job exceeded the maximum time limit for jobs, and has been terminated.`
https://travis-ci.org/Linuxbrew/linuxbrew/builds/119647439#L2036
Succeeded on LinuxbrewTestBot.

Issue #999, opened 2016-03-29
postgresql: Depends on xslt


sjackman:
https://travis-ci.org/Linuxbrew/linuxbrew/builds/119350327#L2610-L2611
```
checking for xsltCleanupGlobals in -lxslt... no
configure: error: library 'xslt' is required for XSLT support
```

rwhogg:
Looks like it depends on `libuuid` too.

```
checking for uuid_generate in -luuid... no
configure: error: library 'uuid' is required for E2FS UUID
```

rwhogg:
I just got it to build successfully, so I'm assigning this to me.

sjackman:
`libuuid` is provided by three packages, which drives me crazy: `libuuid`, `ossp-uid` and `util-linux`.
On Linuxbrew the canonical provider of `libuuid` is `util-linux`, which I believe is actually contains a fork of `libuuid`. I'm thinking that this may need to be abstracted away to a Requirement `:libuuid`.

rwhogg:
> I'm thinking that this may need to be abstracted away to a Requirement `:libuuid`.

Sounds like a plan to me :smile:

I'll put #1019 on hold for now.

sjackman:
In the meantime feel free to use
```ruby
depends_on "util-linux" if OS.linux? # for libuuid
```
That's what I've done in other cases.
Or if you like, take a stab at creating `LibuuidRequirement`.
See https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/requirements.rb

PR #1019, opened 2016-03-31
postgresql: Depends on libxslt and util-linux for Linuxbrew

rwhogg:
Without `libxslt`:
```
checking for xsltCleanupGlobals in -lxslt... no
configure: error: library 'xslt' is required for XSLT support
```

Without `libuuid`:
```
checking for uuid_generate in -luuid... no
configure: error: library 'uuid' is required for E2FS UUID
```

Closes Linuxbrew/linuxbrew#999

### DIFF
--- a/Library/Formula/postgresql.rb
+++ b/Library/Formula/postgresql.rb
@@ -12,7 +12,11 @@ class Postgresql < Formula
 
   option "32-bit"
   option "without-perl", "Build without Perl support"
-  option "without-tcl", "Build without Tcl support"
+  if OS.linux?
+    option "with-tcl", "Build with Tcl support"
+  else
+    option "without-tcl", "Build without Tcl support"
+  end
   option "with-dtrace", "Build with DTrace support"
 
   deprecated_option "no-perl" => "without-perl"
@@ -23,6 +27,9 @@ class Postgresql < Formula
   depends_on "readline"
   depends_on "libxml2" if MacOS.version <= :leopard # Leopard libxml is too old
   depends_on :python => :optional
+  depends_on "libxslt" unless OS.mac?
+  depends_on "util-linux" if OS.linux? # for libuuid
+  depends_on "homebrew/dupes/tcl-tk" if build.with?("tcl") && !OS.mac?
 
   conflicts_with "postgres-xc",
     :because => "postgresql and postgres-xc install the same binaries."


sjackman:
LinuxbrewTestBot:
```
checking for tclsh... no
checking for tcl... no
checking for tclsh8.6... no
checking for tclsh86... no
checking for tclsh8.5... no
checking for tclsh85... no
checking for tclsh8.4... no
checking for tclsh84... no
checking for tclsh8.3... no
checking for tclsh83... no
configure: error: Tcl shell not found
```

rwhogg:
Yeah, there's a `--without-tcl` option, so I'm thinking it should probably do something like

```ruby
depends_on "homebrew/dupes/tcl" if build.with?("tcl") and !OS.mac?
```
We might also need to add something like `--with-x11` or `--without-tk` to prevent it from trying to build the OS X Aqua GUI.

`perl` would have the same problem, but I think we can safely assume the host system provides that by default.

sjackman:
One possible issue is that TCL is included in "homebrew/dupes/tcl-tk", and I believe that depends on X11 unless `--without-x11` is passed.

rwhogg:
> One possible issue is that TCL is included in homebrew/dupes/tcl-tk, and I believe that depends on X11 unless --without-x11 is specified.

There's a few possible ways to solve that:

1. Make `--without-tk` a required argument for `tcl-tk`, as in

```ruby
depends_on "homebrew/dupes/tcl-tk" => ["--without-tk"] if build.with?("tcl") and !OS.mac?
```
 
2. Something similar to what `python` does:

```ruby
depends_on :x11 if build.with?("tcl-tk") && Tab.for_name("homebrew/dupes/tcl-tk").with?("x11")
```

3. Change `--without-tcl` to `--with-tcl` and make the dependency `:optional`:

Of the three, I think option 2 is probably the best (assuming it works; all three of these are untested). If I understand what's happening on https://github.com/Linuxbrew/linuxbrew/issues/1026, then doing that might force people to recompile `tcl-tk` - not ideal. And option 3 would be a factory for merge conflicts, not to mention be a not-strictly-necessary change to the way the formula behaves. I'm sure they had a good reason for wanting `tcl-tk` to be used by default but not `python`, though I'm not really sure what that reason is.

sjackman:
I'd like to make `tcl-tk` not build `tk` by default, and make the option `--with-tk` instead of `--without-tk`, but `tcl-tk` is in `homebrew/dupes`, so I can't modify it greatly.

What do you think of making the default on Linux `--without-tcl`?
```ruby
if OS.linux?
  option "with-tcl", "Build with Tcl support"
else
  option "without-tcl", "Build without Tcl support"
end
```

sjackman:
A tidier solution is to create a `:tcl` Requirement that installs `tcl-tk --without-tk` by default but is satisfied by any installation of `tcl-tk`.

sjackman:
> Make --without-tk a required argument for tcl-tk, as in

The problem with this solution is that `tcl-tk` installed `--with-tk` would be uninstalled to get instead installed `--without-tk`, which is bad.

rwhogg:
> The problem with this solution is that tcl-tk installed --with-tk would be uninstalled to get instead installed --without-tk, which is bad.

Agreed. That would really suck if someone was depending on Tk for some other reason.

`:tcl` sounds like a plan.

rwhogg:
> What do you think of making the default on Linux --without-tcl?

As a temporary workaround to allow `postgres` to build, I like it. As a permanent solution, not a huge fan. `postgres` is an important formula IMO, so in the interest of getting this merged more quickly, I'd be happy to do it.

sjackman:
> `:tcl` sounds like a plan.

Great.

sjackman:
@rwhogg Please merge this patch. It's good to go!

sjackman:
The `:tcl` requirement can wait for another day.

rwhogg:
Sounds good, will merge tonight

On April 5, 2016 8:34:20 PM PDT, Shaun Jackman <notifications@github.com> wrote:
>The `:tcl` requirement can wait for another day.
>
>---
>You are receiving this because you were mentioned.
>https://github.com/Linuxbrew/linuxbrew/pull/1019#issuecomment-206100169

-- 

PR #1023, opened 2016-03-31
ENV/std: ensure LD_LIBRARY_PATH components available

### LINKED ISSUE #1021: `Failed to download resource` while brewing
Yesterday, I ran `brew update` and after that tried `brew upgrade`. At some point all of the formulae were reporting errors such as

```Error: Failed to download resource "brew-caveats"
Failure while executing: git fetch -q origin```

When I added `--verbose` and tried to reproduce the steps everything worked perfectly. A workaround is to `brew fetch formula` then `brew install formula` but even that fails sometimes (eg. for bottled formulae).

Errors are for both git and curl formulae, so it isn't the case of one binary misbehaving.

What is more interesting in `brew-caveats` case after an unsuccessful `brew install` I see `~/.cache/Homebrew/brew-caveats--git` containing `.git` subdir.

Any ideas what could be the problem here?

DoomHammer:
Fixes #1021

### DIFF
--- a/Library/Homebrew/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/ENV/std.rb
@@ -62,7 +62,7 @@ module Stdenv
         # add both lib and prefix to LD_LIBRARY_PATH.
         # segfault when LD_LIBRARY_PATH is set to non-existent directory.
         # See https://github.com/Linuxbrew/linuxbrew/issues/841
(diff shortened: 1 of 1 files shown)


sjackman:
This patch fixes a regression.

sjackman:
:+1: Merge it! Thanks, Piotr.

sjackman:
Please merge! It's good to go.

sjackman:
Thanks, Piotr! Sorry for the regression.

PR #1022, opened 2016-03-31
rtags: fix for linuxbrew

DoomHammer:
Fixes #1018

PR #1024, opened 2016-03-31
test-bot: Don't remove essential deps of Linuxbrew

sjackman:
The essential formula are
`curl git gcc`

Their essential dependencies are
`curl expat glibc gmp isl libmpc mpfr openssl zlib`

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -495,6 +495,14 @@ module Homebrew
         end
       end
 
+      # Do not remove the dependencies of essential formula.
+      essential_formula = OS.linux? ? ["curl", "gcc", "git"] : []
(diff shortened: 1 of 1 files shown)


sjackman:
https://github.com/Linuxbrew/linuxbrew/pull/1022#issuecomment-205927846
> Any plans on merging that one?

@DoomHammer I'd rather teach `test-bot` not to remove dependencies of any installed formula that were installed before it starts its tests. I'd also like to understand why it's removing dependencies of `curl`, which it should know not to do. The `test-bot` doesn't know that `gcc` was sneakily installed without it asking, so I get why that's breaking, but it knows that `curl` was installed.

This patch needs more testing. I'm doing that now.

sjackman:
Resolved by `brew install gcc` before `brew test-bot`.

DoomHammer:
On legacy?

PR #1025, opened 2016-03-31
Add circle.yml

sjackman:


DoomHammer:
So what is the CircleCI status now?

PR #1028, opened 2016-04-01
re2: Fix for Linuxbrew

rwhogg:
There were two problems with `re2`:

1. It was installing libraries with the suffix ".dylib" instead of ".so".
2. It needs C++ 11. (On OS X it gets used by default).

From https://gist.github.com/rwhogg/83095a698e0ce3289c557c02ca1c0eb8#file-01-make-L11 :

```
In file included from util/hash.cc:41:
In file included from ./util/util.h:33:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/atomic:38:
/usr/bin/../lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/bits/c++0x_warning.h:32:2: error: This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
#error This file requires compiler and library support for the \
```

### DIFF
--- a/Library/Formula/re2.rb
+++ b/Library/Formula/re2.rb
@@ -16,11 +16,15 @@ class Re2 < Formula
     sha256 "941571b08e34921134c2fc15b5e855f7c2da5a882fd262766fa01dac988990f3" => :mavericks
   end
 
+  needs :cxx11 unless OS.mac?
+
   def install
+    ENV.cxx11 unless OS.mac?
     system "make", "install", "prefix=#{prefix}"
-    system "install_name_tool", "-id", "#{lib}/libre2.0.dylib", "#{lib}/libre2.0.0.0.dylib"
-    lib.install_symlink "libre2.0.0.0.dylib" => "libre2.0.dylib"
-    lib.install_symlink "libre2.0.0.0.dylib" => "libre2.dylib"
+    system "install_name_tool", "-id", "#{lib}/libre2.0.dylib", "#{lib}/libre2.0.0.0.dylib" if OS.mac?
+    ext = OS.mac? ? "dylib" : "so"
+    lib.install_symlink "libre2.0.0.0.#{ext}" => "libre2.0.#{ext}"
+    lib.install_symlink "libre2.0.0.0.#{ext}" => "libre2.#{ext}"
   end
 
   test do


Library/Formula/re2.rb
```diff
@@ -16,11 +16,15 @@ class Re2 < Formula
     sha256 "941571b08e34921134c2fc15b5e855f7c2da5a882fd262766fa01dac988990f3" => :mavericks
   end
 
+  needs :cxx11 unless OS.mac?
```
sjackman: Do you know why `:cxx11` isn't needed on Mac OS? Is it because it's enabled by default on all the most recent XCode?

rwhogg:
> Error: The selected compiler doesn't support C++11: gcc

Travis is complaining that gcc doesn't support C++11. It's using Ubuntu 12.04, which means GCC should be 4.6, which supports _some_ C++ 11 features.

Will have to dig into this further.

Library/Formula/re2.rb
rwhogg: Yeah, C++ 11 is used by the formula by default on OS X, presumably due to XCode.

sjackman:
I'm okay ignoring the Travis (Ubuntu 12) failure if it builds on LinuxbrewTestBot (Ubuntu 14).

sjackman:
LinuxbrewTestBot says…
```
==> brew install --verbose --build-bottle re2
==> Downloading https://github.com/google/re2/archive/2016-03-01.tar.gz
Already downloaded: /home/travis/.cache/Homebrew/re2-20160301.tar.gz
==> Verifying re2-20160301.tar.gz checksum
tar xf /home/travis/.cache/Homebrew/re2-20160301.tar.gz
Error: The selected compiler doesn't support C++11: gcc
==> FAILED
```
The compiler `gcc` refers to GCC 4.2 in Mac Homebrew land, which is why it thinks that it doesn't support C++11. Ubuntu 14 has GCC 4.8.4, which I believe does support C++11. So, it *should* know that the selected compiler supports C++11. I don't know why it's complaining.

In any case, this PR can be merged as is. We'll solve the C++11 issue another day. Merge it!
:+1:

PR #1031, opened 2016-04-01
Test a formula that requires C++ 11. Do not merge

rwhogg:
This is to test the Travis failure that came up during #1028

### DIFF
--- a/Library/Formula/cppcheck.rb
+++ b/Library/Formula/cppcheck.rb
@@ -1,3 +1,4 @@
+# Test a formula that requires C++ 11. Do not merge.
 class Cppcheck < Formula
   desc "Static analysis of C and C++ code"
   homepage "https://sourceforge.net/projects/cppcheck/"


tseemann:
Regarding `needs :cxx11 unless OS.mac?`

Is there *harm* in having `needs :cxx11` on OSX ?

sjackman:
No, I don't think so. Unless for some reason this formula does *not* require C++11 on Mac OS, but does require C++11 on Linux, which would be odd, but possible. If the default compiler on Mac now supports C++11 by default, then there's going to be lots of places where `needs :cxx11` is accidentally omitted on the Homebrew formula.

The best argument for adding `unless OS.mac?` is so that we can quickly tell from the source code that this line is a Linuxbrew patch.

sjackman:
@rwhogg Close this issue? Both Travis and LinuxbrewTestBot are happy with this build.

Issue #951, opened 2016-03-20
Modify x11_requirement.rb check whether linuxbrew/xorg/xorg is installed


sjackman:
You may want to create a new file `xorg_requirement.rb` and add a line to [`x11_requirement.rb`](https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/requirements/x11_requirement.rb#L26) to switch over to `xorg_requirement` automatically `unless OS.mac?`. Also disable the hack in [xquartz.rb](https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/os/mac/xquartz.rb#L33).

See #527

sjackman:
This issue was moved to Linuxbrew/brew#5

Issue #938, opened 2016-03-15
Error: No available formula with the name "homebrew/dupes/ncurses" 


sjackman:
Installing a formula that depends on `:python` before `homebrew/dupes` is tapped results in the error message `No available formula with the name "homebrew/dupes/ncurses"`. Note that on Linuxbrew `python` depends on `readline`, which depends on `homebrew/dupes/ncurses`. Installing a formula, which depends on `"python"` rather than `:python` works fine. e.g. `brew install gobject-introspection`. See below for examples.

@xu-cheng I believe this behaviour has changed recently, in the last week or so. Any thoughts on what changed?

### Fails

```
❯❯❯ brew install sphinx-doc
Error: No available formula with the name "homebrew/dupes/ncurses" 
Please tap it and then try again: brew tap homebrew/dupes
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```

### Succeeds

```
❯❯❯ brew install gobject-introspection
==> Tapping homebrew/dupes
Cloning into '/home/sjackman/.linuxbrew/Library/Taps/homebrew/homebrew-dupes'...
remote: Counting objects: 42, done.
remote: Compressing objects: 100% (42/42), done.
remote: Total 42 (delta 0), reused 5 (delta 0), pack-reused 0
Unpacking objects: 100% (42/42), done.
Checking connectivity... done.
Tapped 38 formulae (103 files, 105.5K)
==> Installing dependencies for gobject-introspection: xz, pkg-config, li
==> Installing gobject-introspection dependency: xz
==> Downloading https://linuxbrew.bintray.com/bottles/xz-5.2.2.x86_64_linux.bott
Already downloaded: /home/sjackman/.cache/Homebrew/xz-5.2.2.x86_64_linux.bottle.tar.gz
==> Pouring xz-5.2.2.x86_64_linux.bottle.tar.gz
…
```

xu-cheng:
Could you submit this issue to Homebrew? This is just one of many bugs in our fragile dependency system.

sjackman:
Fixed by @ilovezfs in https://github.com/Homebrew/legacy-homebrew/issues/50271

PR #1039, opened 2016-04-04
Create xorg_requirement.rb

rwhogg:
This creates a new `XorgRequirement` class and hacks `X11Requirement` to defer to it as necessary.
It also removes some hackery from xquartz.rb that just pretended X11 was already installed.

Closes Linuxbrew/linuxbrew#951

- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your core changes, as applicable? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031) if you'd like one.
- [ ] Have you successfully ran `brew tests` with your changes locally?

### DIFF
--- a/Library/Homebrew/os/mac/xquartz.rb
+++ b/Library/Homebrew/os/mac/xquartz.rb
@@ -30,7 +30,6 @@ module OS
       # The X11.app distributed by Apple is also XQuartz, and therefore covered
       # by this method.
       def version
-        @version ||= latest_version if OS.linux?
         @version ||= detect_version
       end
 
@@ -95,7 +94,6 @@ module OS
       # remain public. New code should use MacOS::X11.bin, MacOS::X11.lib and
       # MacOS::X11.include instead, as that accounts for Xcode-only systems.
       def prefix
-        @prefix ||= Pathname.new("/usr") if OS.linux?
         @prefix ||= if Pathname.new("/opt/X11/lib/libpng.dylib").exist?
           Pathname.new("/opt/X11")
         elsif Pathname.new("/usr/X11/lib/libpng.dylib").exist?
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -94,7 +94,7 @@ class DependencyCollector
 
   def parse_symbol_spec(spec, tags)
     case spec
-    when :x11        then X11Requirement.new(spec.to_s, tags)
+    when :x11        then OS.mac? ? X11Requirement.new(spec.to_s, tags) : XorgRequirement.new(tags)
     when :xcode      then XcodeRequirement.new(tags)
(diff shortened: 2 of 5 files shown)


rwhogg:
I cannot emphasize enough how much a work-in-progress this is. It works, but telling the various `X11Requirement` methods to defer to their `XorgRequirement` counterparts is just awful.

What I think might work better is having `XorgRequirement` extend `X11Requirement` and mapping `depends_on :x11` to 

Nevertheless, submitting the PR so there's a visible indication of my progress.

Library/Homebrew/os/mac/xquartz.rb
```diff
@@ -95,7 +94,6 @@ def provided_by_apple?
       # remain public. New code should use MacOS::X11.bin, MacOS::X11.lib and
       # MacOS::X11.include instead, as that accounts for Xcode-only systems.
       def prefix
-        @prefix ||= Pathname.new("/usr") if OS.linux?
```
rwhogg: I've just removed this for now, but I think I need to do more than that.

Assuming we want to prefer the Linuxbrew/xorg/xorg installation over anything else, something like this might make more sense (pardon my mediocre Ruby-fu):

```ruby
if OS.linux?
  @prefix ||= if Pathname.new(cellar/"xorg").exist?
    # Not sure if this case is really necessary... not all that familiar with this code ATM
    Pathname.new(ENV["HOMEBREW_PREFIX"]
  elsif Pathname.new("/usr/lib/x86_64-linux-gnu/libX11.so").exist?
    Pathname.new("/usr")
  end
end
```

tseemann: On RHEL/Centos that `libX11.so` doesn't exist - it is in `/usr/lib64/libX11.so`

sjackman:
LinuxbrewTestBot error:
```
  1) Error:
X11RequirementTests#test_satisfied:
TapFormulaUnavailableError: No available formula with the name "linuxbrew/xorg/xorg" 
Please tap it and then try again: brew tap linuxbrew/xorg
```
I thought that this bug was fixed. =/
See Linuxbrew/linuxbrew#938 and Homebrew/legacy-homebrew#50271

sjackman:
> What I think might work better is having `XorgRequirement` extend `X11Requirement` and mapping `depends_on :x11` to `XorgRequirement`

:+1:

sjackman:
```
$ brew install gobject-introspection
==> Tapping homebrew/dupes
Cloning into '/home/sjackman/.linuxbrew/Library/Taps/homebrew/homebrew-dupes'...
```
It looks fixed to me. I wonder why this requirement doesn't tap. Perhaps because it's in `brew tests` which behaviours differently than reality some times…

rwhogg:
Moved to https://github.com/Linuxbrew/brew/pull/10

PR #1043, opened 2016-04-04
node: Do not merge. Build a bottle for Linuxbrew

sjackman:


sjackman:
`brew test-bot --keep-old took more than 10 minutes since last output`

Issue #1057, opened 2016-04-08
go: unknown relocation type 42; compiled without -fpic?


xu-cheng:
_From @int03h on April 8, 2016 15:16_

Trying to brew install go on debian SID .. (I did try it on testing too) .. it does compile natively on the OS, not problem. ( tried with and without export `HOMEBREW_BUILD_FROM_SOURCE=1`) 

I compiled ethereum-cpp successfully but I can't get GETH ( go version to compile because of it's dependency on golang ) 
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [ ] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

https://gist.github.com/1b3367f01fd05e67af2c027782cab3f9

```
/tmp/go20160408-29305-1dninkg/go/gobootstrap/pkg/linux_amd64/runtime/cgo.a(_all.o): unknown relocation type 42; compiled without -fpic?
/tmp/go20160408-29305-1dninkg/go/gobootstrap/pkg/linux_amd64/runtime/cgo.a(_all.o): unknown relocation type 42; compiled without -fpic?
runtime/cgo(.text): unexpected relocation type 298
runtime/cgo(.text): unexpected relocation type 298
# cmd/go
/tmp/go20160408-29305-1dninkg/go/gobootstrap/pkg/linux_amd64/runtime/cgo.a(_all.o): unknown relocation type 42; compiled without -fpic?
/tmp/go20160408-29305-1dninkg/go/gobootstrap/pkg/linux_amd64/runtime/cgo.a(_all.o): unknown relocation type 42; compiled without -fpic?
runtime/cgo(.text): unexpected relocation type 298
runtime/cgo(.text): unexpected relocation type 298
```

thanks! 
int

_Copied from original issue: Homebrew/brew#48_

sjackman:
I have seen other issues with this error message reported. Try
`brew install binutils go`
I'm not convinced that will work, but it's worth a shot.

sjackman:
It looks as though this issue is known upstream. See
https://github.com/golang/go/issues/13114
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=807136

sjackman:
Maybe my `binutils` guess is correct. See https://github.com/Linuxbrew/linuxbrew/issues/1057

[...]

PR #1058, opened 2016-04-08
go: Fix unknown relocation type 42 for Linuxbrew

sjackman:
Closes Linuxbrew/linuxbrew#1057.

### DIFF
--- a/Library/Formula/go.rb
+++ b/Library/Formula/go.rb
@@ -46,6 +46,10 @@ class Go < Formula
   end
 
   def install
+    # Fix error: unknown relocation type 42; compiled without -fpic?
+    # See https://github.com/Linuxbrew/linuxbrew/issues/1057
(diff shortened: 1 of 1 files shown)


sjackman:
Unfortunately bottling failed due to a static executable. I have to figure out how to distinguish static ELF executables from dynamic ones. Perhaps using `libmagic`, or a Ruby ELF library.
```
==> brew bottle --verbose --rb go --keep-old
==> Bottling go-1.6.x86_64_linux.bottle.2.tar.gz...
/home/linuxbrew/.linuxbrew/Cellar/patchelf/0.9/bin/patchelf --set-rpath @@HOMEBREW_PREFIX@@/lib --set-interpreter /lib64/ld-linux-x86-64.so.2 /home/linuxbrew/.linuxbrew/Cellar/go/1.6/libexec/bin/go 
cannot find section 
Error: undefined method `write' for nil:NilClass
```

Issue #1053, opened 2016-04-06
[openssl] failed on Ubuntu 14.04.4 LTS


channprj:
Thank you for any help in advance. I closely examine the issues, but I can't find solution.

I'm going to install pyenv or pip, but brew doesn't work.
(I installed it another way, but I want to use it on brew.)

Logs are here:

```shell
==> Downloading https://www.openssl.org/source/openssl-1.0.2g.tar.gz
Already downloaded: /home/channprj/.cache/Homebrew/openssl-1.0.2g.tar.gz
==> Downloading https://raw.githubusercontent.com/Homebrew/patches/3f1dc8ea145a70543aded8101a0c725abf82fc45/openssl/revert-pass-pure-constants-verbatim.patch
Already downloaded: /home/channprj/.cache/Homebrew/openssl--patch-e38f84181a56e70028ade8408ad70aaffaea386b7e1b35de55728ae878d544aa.patch
==> Downloading https://raw.githubusercontent.com/Homebrew/patches/3f1dc8ea145a70543aded8101a0c725abf82fc45/openssl/tshort-asm.patch
Already downloaded: /home/channprj/.cache/Homebrew/openssl--patch-f161e2fc1395efcb53d785004d67d4962d28aa8ce282a91020f12809c03b2afd.patch
==> Patching
==> Applying revert-pass-pure-constants-verbatim.patch
patching file crypto/perlasm/x86_64-xlate.pl
==> Applying tshort-asm.patch
patching file crypto/perlasm/x86_64-xlate.pl
==> perl ./Configure --prefix=/home/channprj/.linuxbrew/Cellar/openssl/1.0.2g --openssldir=/home/channprj/.linuxbrew/etc/openssl no-ssl2 zlib-dynamic shared enable-cms -isystem/home/channprj/.linuxbrew/include -Os -w -pipe -march=core2 -L/home/channprj/.linuxbrew/lib -Wl,--dynamic-linker=/home/channprj/.linuxbrew/lib/ld.so -Wl,-rpath,/home/channprj/.linuxbrew/lib linux-x86_64
==> make depend
==> make
==> make test
Last 15 lines from /home/channprj/.cache/Homebrew/Logs/openssl/04.make:
test sslv2
Testing was requested for a disabled protocol. Skipping tests.
test sslv2 with server authentication
Testing was requested for a disabled protocol. Skipping tests.
test sslv2 with client authentication
Testing was requested for a disabled protocol. Skipping tests.
[...]

PR #1094, opened 2016-04-21
core: Fix march for linux

alekseysidorov:
### Changes to Linuxbrew's Core:

- [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your core changes, as applicable? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031) if you'd like one.
- [ ] Have you successfully ran `brew tests` with your changes locally?

### DIFF
--- a/Library/Homebrew/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/ENV/std.rb
@@ -8,7 +8,7 @@ module Stdenv
 
   # @private
   SAFE_CFLAGS_FLAGS = "-w -pipe"
-  DEFAULT_FLAGS = "-march=core2 -msse4"
+  DEFAULT_FLAGS = OS.mac? ? "-march=core2 -msse4" : "-march=native"
 
   def self.extended(base)
     unless ORIGINAL_PATHS.include? HOMEBREW_PREFIX/"bin"
@@ -361,7 +361,7 @@ module Stdenv
   def effective_arch
     if ARGV.build_bottle?
       ARGV.bottle_arch || Hardware.oldest_cpu
-    elsif Hardware::CPU.intel? && !Hardware::CPU.sse4?
+    elsif OS.mac? && Hardware::CPU.intel? && !Hardware::CPU.sse4?
       # If the CPU doesn't support SSE4, we cannot trust -march=native or
       # -march=<cpu family> to do the right thing because we might be running
       # in a VM or on a Hackintosh.


sjackman:
Fixes Linuxbrew/linuxbrew#1053

sjackman:
I'm currently splitting `Linuxbrew/linuxbrew` into `Linuxbrew/brew` and `Linuxbrew/homebrew-core`, as Homebrew did last week (two weeks ago?), so I'm not merging any pull requests right now. I'll cherry-pick this PR after the split is complete, probably next week. Sorry for the delay.

sjackman:
👍 Look's great! Thanks, Aleksey.

eiennohito:
What should I do if I have a grid of machines with different CPU and want to specify a the oldest march, native won't work for me.

sjackman:
If you specify `--build-bottle` it will target `-march=core2`. Will that work for you?

eiennohito:
The ability to override target architecture using an environment variable would be better and less error prone. I definitely will forget about bottles myself some time in future and I would need to teach other people to use that.

sjackman:
`brew install --build-bottle foo` installs a formula as normal but sets `-march=core2` and disables postinstall. There's also the `--bottle-arch` option, but the value has to be from one of the values found in
https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/os/linux/hardware.rb#L3-L5

eiennohito:
Is it possible to make Homebrew prefer value of `-march` in `CFLAGS`/`CXXFLAGS` to guessed ones?

sjackman:
No, `CFLAGS` and `CXXFLAGS` and other environment variables are intentionally removed from the environment before building for reproducibility. What architecture are you targeting? i.e. what value of `-march` do you want to use?

eiennohito:
We have Xeons from Nehalem and up. So core2.

By the way, will the --build-bottle flag work with upgrade command?

sjackman:
A `HOMEBREW_ARCH` environment variable may be a better fit for your use case. Can you please open an feature request issue to discuss?

eiennohito:
Okay. Should I do it in this repo?

sjackman:
Yes, please.

alekseysidorov:
When you are going to cherry-pick this fix?

sjackman:
Sorry for the delay, and thanks for the bump, Aleksey. I'll do it today.

sjackman:
Moved to https://github.com/Linuxbrew/brew/pull/21

PR #1103, opened 2016-04-30
vim v7.4.1799

gutierri:


rwhogg:
Hey @gutierri : 

1. Please send version updates to Homebrew, not Linuxbrew, unless it's a Linux-only formula. (You can usually tell by the presence of the text `# tag "linuxbrew"`). We'll get the change post-merge.
3. We have split this repository into Linuxbrew/brew (Homebrew commands and libraries) and Linuxbrew/homebrew-core (formulae). Please send version updates and new formulae to the homebrew-core repository for the appropriate project from now on (this is true for both Homebrew and Linuxbrew). We're still working on fully migrating everyone to the split repos; I apologize for the growing pains.
3. In this particular instance, I don't think upstream will merge the PR since they have a specific schedule for updating vim.

rwhogg:
Closing based on points 1 and 3. Thanks for contributing!

retokromer:
@rwhogg Just in case you haven't noticed it, this could also be useful https://github.com/Homebrew/brew/pull/168 which would be a big step forward, in my opinion.

PR #1126, opened 2017-09-28
Troubleshooting: update link.

keikun17:
brew

still leads to the `legacy-linuxbrew` project so i figured it this might still  be the right project to put this PR.


maxim-belkin:
Thank you, @keikun17!

sjackman:
Note that `Linuxbrew/linuxbrew` is a very old repository, replaced by `Linuxbrew/brew`. You'll want to use the new installation instructions at…
http://linuxbrew.sh/#install-linuxbrew

